PR #6, opened 2016-03-15
Allow all users to send mails to the name list in etc/mail_sysop that…

sitos:
… could be configured in xfile.

### DIFF
--- a/include/proto.h
+++ b/include/proto.h
@@ -399,6 +399,7 @@ int setforward(void);
 int mail_mbox(void);
 int built_mail_index(void);
 int mail_all(void);
+int mail_account_sysop(void);
 int invalidaddr(const char *addr);
 int is_valid_email(const char *full_address);
 int do_send(const char *userid, const char *title, const char *log_source);
--- a/mbbsd/mail.c
+++ b/mbbsd/mail.c
@@ -1195,6 +1195,103 @@ mail_all(void)
 }
 
 int
+get_account_sysop(struct Vector *namelist)
+{
+    FILE *fp;
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP, "r")) != NULL) {
+        char line[STRLEN+1];
+        while(fgets(line, sizeof(line), fp)) {
+            char userid[IDLEN+1];
+
+            chomp(line);
+            if (is_validuserid(line) && searchuser(line, userid))
+                Vector_add(namelist, line);
+        }
+        fclose(fp);
+        return 0;
+    } else
+        return -1;
+}
+
+
+int
+mail_account_sysop(void)
+{
+    // send mail to account sysop
+    int             i;
+    fileheader_t    mymail;
+    char            fpath[TTLEN];
+    char            genbuf[PATHLEN];
+    char            idbuf[IDLEN+1];
+    int             oldstat = currstat;
+    char            save_title[STRLEN];
+    char            tmp_title[STRLEN-20];
+    struct          Vector namelist;
+
+    Vector_init(&namelist, IDLEN+1);
+
+    if (get_account_sysop(&namelist) == -1)
+    {
+        vmsg("讀取收件人名單失敗，請向 " BN_BUGREPORT " 回報");
+        Vector_delete(&namelist);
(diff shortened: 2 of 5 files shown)


mbbsd/mail.c
```diff
@@ -1194,6 +1194,101 @@ mail_all(void)
     return 0;
 }
 
+void
+get_account_manager(struct Vector *namelist)
+{
+    FILE *fp;
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP, "r")) != NULL) {
+        char line[STRLEN+1];
+        while(fgets(line, STRLEN + 1, fp)) {
```
kcwu: s/STRLEN+1/sizeof(line)/

mbbsd/mail.c
```diff
@@ -1194,6 +1194,101 @@ mail_all(void)
...
+            char *ptr = strstr(line, "\n");
```
kcwu: fgets guarantee '\0' in the buffer. So these 4 lines could be replaced by
chomp(line);

https://github.com/ptt/pttbbs/blob/master/common/sys/string.c#L67

mbbsd/mail.c
```diff
@@ -1194,6 +1194,101 @@ mail_all(void)
...
+    char            idbuf[IDLEN+1];
+    int             oldstat = currstat;
+    char            save_title[STRLEN];
+
+    struct Vector namelist;
+    Vector_init(&namelist, IDLEN+1);
+
+    get_account_manager(&namelist);
+
+    if (Vector_length(&namelist) == 0)
```
robertabcd, line 1240: I think it's better to distinguish between no namelist and failed to load.

mbbsd/mail.c
```diff
@@ -1194,6 +1194,101 @@ mail_all(void)
     return 0;
 }
 
+void
+get_account_manager(struct Vector *namelist)
```
robertabcd: There are two namings in this commit: account manager, and account sysop. Probably sticking to one of them would be better.

robertabcd:
Oh, please set the author to a reachable email address. (currently is sitos@node2.cluster)

include/proto.h
```diff
@@ -399,6 +399,7 @@ int setforward(void);
 int mail_mbox(void);
 int built_mail_index(void);
 int mail_all(void);
+int mail_am(void);
```
hungte: I think we prefer a readable name, for example

 mail_account_sysop

mbbsd/mail.c
```diff
@@ -1195,6 +1195,105 @@ mail_all(void)
...
 int
+get_account_sysop(struct Vector *namelist)
+{
+    FILE *fp;
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP, "r")) != NULL) {
+        char line[STRLEN+1];
+        while(fgets(line, sizeof(line), fp)) {
+            chomp(line);
+
+            char userid[IDLEN+1];
```
hungte, line 1204: I'd prefer to not declaring variables after code.

Try to move this some where after char line.

mbbsd/mail.c
```diff
@@ -1195,6 +1195,105 @@ mail_all(void)
...
+
+    hold_mail(fpath, NULL, save_title);
+    unlink(fpath);
+
+    setutmpmode(oldstat);
+    clear();
+
+    pressanykey();
+
+    Vector_delete(&namelist);
```
hungte: We should delete the namelist as early as possible, especially before that pressanykey() since that may go into recursive call.

mbbsd/mail.c
```diff
@@ -1195,6 +1195,105 @@ mail_all(void)
...
+        if (append_record_forward(genbuf, &mymail, sizeof(mymail), idbuf) == -1)
+            vmsg(err_uid);
+        sendalert(idbuf, ALERT_NEW_MAIL);
+    }
+
+    hold_mail(fpath, NULL, save_title);
+    unlink(fpath);
+
+    setutmpmode(oldstat);
+    clear();
```
hungte: Do you really want to clear? I think a pressanykey is good enough

PR #7, opened 2016-04-01
1. Add the description of the function for sending mails to account s…

sitos:
…ysops
1. Allow the users to select an account sysop to send the mail

### DIFF
--- a/mbbsd/mail.c
+++ b/mbbsd/mail.c
@@ -1213,41 +1213,91 @@ get_account_sysop(struct Vector *namelist)
         return -1;
 }
 
+void
+show_mail_account_sysop_desc(void)
+{
+    more(FN_MAIL_ACCOUNT_SYSOP_DESC, 1);
+}
+
+const char *
+select_account_sysop(struct Vector *namelist)
+{
+    int i, sysop_index;
+    char linebuf[STRLEN+1];
+
+    vs_hdr("請選擇要寄給哪一位帳號站長");
+    for (i = 0; i < Vector_length(namelist) && i + 2 < t_lines; i++) {
+        const char *userid = Vector_get(namelist, i);
+        snprintf(linebuf, sizeof(linebuf), "%2d. %s", i + 1, userid);
+        move(i + 2, 0);
+        outs(linebuf);
+    }
+    getdata(1, 0, "請輸入帳號站長編號: ", linebuf, 4, LCECHO);
+
+    // This implementation allows users to select the account sysop
+    // who is not showed on the screen due to the limitation of screen size
+    if (sscanf(linebuf, "%d", &sysop_index) != 1 || sysop_index < 1 || sysop_index > Vector_length(namelist))
+        return NULL;
+
+    return Vector_get(namelist, sysop_index - 1);
+}
 
 int
 mail_account_sysop(void)
 {
     // send mail to account sysop
-    int             i;
     fileheader_t    mymail;
     char            fpath[TTLEN];
     char            genbuf[PATHLEN];
     char            idbuf[IDLEN+1];
     int             oldstat = currstat;
     char            save_title[STRLEN];
-    char            tmp_title[STRLEN-20];
     struct          Vector namelist;
+    char            linebuf[STRLEN+1];
+    const char     *userid;
 
+    setutmpmode(SMAIL);
     Vector_init(&namelist, IDLEN+1);
 
     if (get_account_sysop(&namelist) == -1)
     {
-        vmsg("讀取收件人名單失敗，請向 " BN_BUGREPORT " 回報");
         Vector_delete(&namelist);
+        vmsg("讀取收件人名單失敗，請向 " BN_SYSOP " 回報");
         return DIRCHANGED;
     }
 
     if (Vector_length(&namelist) == 0)
     {
-        vmsg("無有效收件人名單，請向 " BN_BUGREPORT " 回報");
         Vector_delete(&namelist);
+        vmsg("無有效收件人名單，請向 " BN_SYSOP " 回報");
         return DIRCHANGED;
     }
 
-    getdata(2, 0, "主題:", tmp_title, sizeof(tmp_title), DOECHO);
-    strlcpy(save_title, tmp_title, sizeof(save_title));
+    clear();
+    show_mail_account_sysop_desc();
 
-    setutmpmode(SMAIL);
+    userid = select_account_sysop(&namelist);
+    Vector_delete(&namelist);
+    if (userid == NULL)
+    {
+        setutmpmode(oldstat);
+        vmsg("輸入編號不正確");
+        return DIRCHANGED;
+    }
+
+    if(!searchuser(userid, idbuf))
+    {
+        setutmpmode(oldstat);
+        vmsg("選取的收件人無效，請向 " BN_SYSOP " 回報");
+        return DIRCHANGED;
+    }
+ 
+    clear();
+    vs_hdr("寄信給帳號站長");
+    snprintf(linebuf, sizeof(linebuf), "寄給 %s 站長", idbuf);
+    move(1, 0);
+    outs(linebuf);
+    getdata(2, 0, "主題:", save_title, STRLEN - strlen("主題:"), DOECHO);
 
     setuserfile(fpath, fn_notes);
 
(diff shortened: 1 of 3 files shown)


mbbsd/mail.c
```diff
@@ -1213,6 +1213,32 @@ get_account_sysop(struct Vector *namelist)
...
+show_mail_account_sysop_desc(void)
+{
+    int index = 1;
+    FILE *fp;
+
+    vs_hdr("�b������H�c����");
+
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP_DESC, "r")) != NULL) {
+        char line[STRLEN+1];
+        while(fgets(line, sizeof(line), fp)) {
```
robertabcd: index is unused in the else block. could you move it here?
for (int i = 0; i < 24 && fgets(...) != NULL; i++) {

mbbsd/mail.c
```diff
@@ -1213,6 +1213,32 @@ get_account_sysop(struct Vector *namelist)
...
+            chomp(line);
+            move(index, 0);
+            outs(line);
+            index++;
+            if (index == 24)
+                break;
+        }
+        fclose(fp);
+        return;
```
robertabcd: This seems to be unnecessary.

mbbsd/mail.c
```diff
@@ -1213,6 +1213,32 @@ get_account_sysop(struct Vector *namelist)
...
+    } else {
+        move(1, 0);
+        outs("�ثe�|�L����");
+        return;
```
robertabcd: Same here. Does it really make sense to show "目前尚無說明"?

mbbsd/mail.c
```diff
@@ -1244,6 +1272,39 @@ mail_account_sysop(void)
...
+        return DIRCHANGED;
+    }
+    sscanf(linebuf, "%d", &sysop_index);
+    if (sysop_index < 1 || sysop_index > Vector_length(&namelist))
+    {
+        vmsg("��J�s�������T");
+        Vector_delete(&namelist);
+        return DIRCHANGED;
+    }
+    sysop_index--;
```
robertabcd: Logic of selecting account sysop makes this function too long. Could you move it to another function and make here just:
if (select_account_sysop(userid) < 0) { /\* error handling */ }
And those cleanups in later code would be simpler as well.

mbbsd/mail.c
```diff
@@ -1227,6 +1253,8 @@ mail_account_sysop(void)
     char            save_title[STRLEN];
     char            tmp_title[STRLEN-20];
     struct          Vector namelist;
+    int             sysop_index = 0;
+    char            linebuf[STRLEN+1];
```
robertabcd, line 1256: Do setutmpmode(SMAIL); before handling any user inputs.

mbbsd/mail.c
robertabcd: Same here.

robertabcd: And does it make sense to show "目前尚無說明"?

mbbsd/mail.c
```diff
@@ -1244,6 +1272,39 @@ mail_account_sysop(void)
...
+        Vector_delete(&namelist);
+        return DIRCHANGED;
+    }
+    sscanf(linebuf, "%d", &sysop_index);
+    if (sysop_index < 1 || sysop_index > Vector_length(&namelist))
+    {
+        vmsg("��J�s�������T");
+        Vector_delete(&namelist);
+        return DIRCHANGED;
+    }
```
robertabcd, line 1293: Could you make these together so the code can be shorter?
if (sscanf(...) != 1 || index < 1 || index > Vector_length(...)) { /\* error */ }
The implementation is slightly different (accepts &lt;space&gt;1) but I think the original check (only first char) is not adequate either.

mbbsd/mail.c
```diff
@@ -1213,6 +1213,32 @@ get_account_sysop(struct Vector *namelist)
...
+    vs_hdr("�b������H�c����");
+
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP_DESC, "r")) != NULL) {
+        char line[STRLEN+1];
+        while(fgets(line, sizeof(line), fp)) {
+            chomp(line);
+            move(index, 0);
+            outs(line);
+            index++;
+            if (index == 24)
```
kcwu: avoid magic number 24.
Please find its corresponding named constant.

mbbsd/mail.c
```diff
@@ -1244,6 +1272,39 @@ mail_account_sysop(void)
...
+    clear();
+    vs_hdr("�п�ܭn�H�����@��b������"); 
+    for (i = 0; i < Vector_length(&namelist) && i + 2 < 24; i++) {
+        const char *userid = Vector_get(&namelist, i);
+        snprintf(linebuf, sizeof(linebuf), "%2d. %s", i + 1, userid);
+        move(i + 2, 0);
+        outs(linebuf);
+    }
+    getdata(1, 0, "�п�J�b������s��: ", linebuf, 4, LCECHO);
+    if (linebuf[0] < '1' || linebuf[0] > '9')
```
kcwu: This check is unnecessary since you check its value at line 1295.

mbbsd/mail.c
```diff
@@ -1261,24 +1322,29 @@ mail_account_sysop(void)
...
-
-        if (append_record_forward(genbuf, &mymail, sizeof(mymail), idbuf) == -1)
-            vmsg(err_uid);
-        sendalert(idbuf, ALERT_NEW_MAIL);
+    const char *userid = Vector_get(&namelist, sysop_index);
+    if(!searchuser(userid, idbuf))
+    {
+        unlink(fpath);
+        setutmpmode(oldstat);
+        vmsg("�����H�L�ġA�ЦV " BN_BUGREPORT " �^��");
```
kcwu: This seems not "bug" for BN_BUGREPORT. If sysop list is invalid, it should be reported to sysop board.

mbbsd/mail.c
```diff
@@ -1213,6 +1213,32 @@ get_account_sysop(struct Vector *namelist)
...
 
+void
+show_mail_account_sysop_desc(void)
+{
+    int index = 1;
+    FILE *fp;
+
+    vs_hdr("�b������H�c����");
+
+    if ( (fp = fopen(FN_MAIL_ACCOUNT_SYSOP_DESC, "r")) != NULL) {
```
kcwu: Just curious, why not use pmore() ?

mbbsd/mail.c
```diff
@@ -1227,6 +1253,8 @@ mail_account_sysop(void)
     char            save_title[STRLEN];
     char            tmp_title[STRLEN-20];
```
kcwu: I forgot why STRLEN-20?
1. why not STRLEN? I am nervous to read that and afraid STRLEN<20. 
2. what is "20"?

mbbsd/mail.c
```diff
@@ -1261,24 +1322,29 @@ mail_account_sysop(void)
...
-        strlcpy(mymail.owner, cuser.userid, sizeof(mymail.owner));
-        strlcpy(mymail.title, save_title, sizeof(mymail.title));
-        mymail.filemode |= FILE_MULTI;
-        sethomedir(genbuf, idbuf);
-
-        if (append_record_forward(genbuf, &mymail, sizeof(mymail), idbuf) == -1)
-            vmsg(err_uid);
-        sendalert(idbuf, ALERT_NEW_MAIL);
+    const char *userid = Vector_get(&namelist, sysop_index);
+    if(!searchuser(userid, idbuf))
```
kcwu: A minor issue: it's bad user expierence that the user already wrote a lengthy mail and you told him the recipient is wrong and the mail is gone.

mbbsd/mail.c
sitos: This implementation is copied&pasted from the function "do_send", but I don't know why STRLEN-20 and what is 20.

mbbsd/mail.c
```diff
@@ -1213,42 +1213,94 @@ get_account_sysop(struct Vector *namelist)
...
+}
+
+const char *
+select_account_sysop(struct Vector *namelist)
+{
+    int i, sysop_index;
+    char linebuf[STRLEN+1];
+
+    vs_hdr("�п�ܭn�H�����@��b������");
+    for (i = 0; i < Vector_length(namelist) && i + 2 < t_lines; i++) {
```
kcwu, line 1229: 這邊有個小問題, 列表太長的話, 看不到但還是可以選.
不過我也沒有簡單的解法, 如果是我的話, 我也懶得處理, 不過至少加註解說這邊有問題.

mbbsd/mail.c
```diff
@@ -1261,24 +1313,20 @@ mail_account_sysop(void)
...
+    mymail.filemode |= FILE_MULTI;
+    sethomedir(genbuf, idbuf);
 
-        strlcpy(mymail.owner, cuser.userid, sizeof(mymail.owner));
-        strlcpy(mymail.title, save_title, sizeof(mymail.title));
-        mymail.filemode |= FILE_MULTI;
-        sethomedir(genbuf, idbuf);
+    if (append_record_forward(genbuf, &mymail, sizeof(mymail), idbuf) == -1)
+        vmsg(err_uid);
+    sendalert(idbuf, ALERT_NEW_MAIL);
```
kcwu: I'm not sure. Do we still need to sendalert() if append_record_forward() failed?

mbbsd/mail.c
```diff
@@ -1213,42 +1213,96 @@ get_account_sysop(struct Vector *namelist)
...
+        return DIRCHANGED;
+    }
+ 
+    clear();
+    vs_hdr("�H�H���b������");
+    snprintf(linebuf, sizeof(linebuf), "�H�� %s ����", idbuf);
+    move(1, 0);
+    outs(linebuf);
+    getdata(2, 0, "�D�D:", tmp_title, STRLEN - strlen("�D�D:"), DOECHO);
+    strlcpy(save_title, tmp_title, sizeof(save_title));
```
robertabcd: Do we really need this tmp_title?

mbbsd/mail.c
```diff
@@ -1213,42 +1213,96 @@ get_account_sysop(struct Vector *namelist)
...
         Vector_delete(&namelist);
         return DIRCHANGED;
     }
 
-    getdata(2, 0, "�D�D:", tmp_title, sizeof(tmp_title), DOECHO);
-    strlcpy(save_title, tmp_title, sizeof(save_title));
+    clear();
+    show_mail_account_sysop_desc();
 
     setutmpmode(SMAIL);
```
robertabcd: I suggest moving this above just after we keep the currstat.
And we could optimize these select sysop logic into another function, then here can just:
if (get_account_sysop(&namelist) == -1 || select_account_sysop(&namelist, idbuf) < 0) { ... }

mbbsd/mail.c
```diff
@@ -1261,27 +1315,27 @@ mail_account_sysop(void)
...
+    }
+    else
+    {
         sendalert(idbuf, ALERT_NEW_MAIL);
+        hold_mail(fpath, NULL, save_title);
     }
-    Vector_delete(&namelist);
 
-    hold_mail(fpath, NULL, save_title);
+    Vector_delete(&namelist);
```
robertabcd: Please destruct it as soon as it's no longer needed.

mbbsd/mail.c
sitos: I think that is probably not a good idea to put the error handling of "get_account_sysop" and "select_account_sysop" together, because the former is about the error of getting the ID list of account sysops that should be reported, but the latter is about the error of selecting an invalid sequence number of account sysops that should be corrected by the user himself. So, the error messages of these two situations should be different.

mbbsd/mail.c
```diff
@@ -1213,41 +1213,91 @@ get_account_sysop(struct Vector *namelist)
...
     }
 
-    getdata(2, 0, "�D�D:", tmp_title, sizeof(tmp_title), DOECHO);
-    strlcpy(save_title, tmp_title, sizeof(save_title));
+    clear();
+    show_mail_account_sysop_desc();
 
-    setutmpmode(SMAIL);
+    userid = select_account_sysop(&namelist);
+    Vector_delete(&namelist);
```
kcwu, line 1280: you delete the list too early.
After deleted, userid become dangling pointer.

hungte:
Dead PR, close now.

PR #8, opened 2016-06-17
Remove obsolete innd which causes infinite loop.

ShikChen:
```
$ bbsadm@pttbbs:~/pttbbs/sample$ pmake install
# some output...
/bin/sh: 1: cd: can't cd to innd
# error in pmake and infinite loop...
```

### DIFF
--- a/sample/Makefile
+++ b/sample/Makefile
@@ -1,4 +1,4 @@
-SUBDIR=etc innd
+SUBDIR=etc
 BBSHOME?=$(HOME)
 
 all:


ShikChen:
Note that there is an `innd` in [util/dailybackup.pl:41](https://github.com/ptt/pttbbs/blob/34db4d2b567e1ea6a6f76950f2c258d8cf3b4c99/util/dailybackup.pl#L41), too.
But I don't know whether you want to preserve it or not 😸

robertabcd:
Thanks.

PR #10, opened 2016-06-21
C++11 style code (noexcept, forwarding reference, move), and spurious wakeup

Fdhvdu:
you may need a semaphore to prevent spurious wakeup

### DIFF
--- a/daemon/boardd/queue.hpp
+++ b/daemon/boardd/queue.hpp
@@ -2,64 +2,67 @@
 // The MIT License.
 #ifndef _QUEUE_H_
 #define _QUEUE_H_
-#include <queue>
-#include <mutex>
-#include <condition_variable>
+#include<condition_variable>
+#include<mutex>
+#include<queue>
+#include<stdexcept> //runtime_error
+#include<utility> //forward, move
 
-template <class T>
+template<class T>
 class Queue
 {
+	std::queue<T> qu_;
+	std::mutex mu_;
+	std::condition_variable cv_pushed_;
+	std::condition_variable cv_poped_;
+	bool eof_;
+	std::size_t max_;
 public:
-  Queue(size_t max = 0)
-    : eof_(false)
-    , max_(max)
-  {}
-
-  void Push(T elem)
-  {
-    std::unique_lock<std::mutex> lk(mu_);
-    while (true) {
-      if (eof_)
-        throw std::runtime_error("Queue closed");
-      if (max_ && qu_.size() >= max_) {
-        cv_poped_.wait(lk);
-        continue;
-      }
-      qu_.push(elem);
-      cv_pushed_.notify_one();
-      return;
-    }
-  }
-
-  bool Pop(T &elem)
-  {
-    std::unique_lock<std::mutex> lk(mu_);
-    while (qu_.empty()) {
-      if (eof_)
-        return false;
-      cv_pushed_.wait(lk);
-    }
-    elem = qu_.front();
-    qu_.pop();
-    cv_poped_.notify_one();
-    return true;
-  }
-
-  void Close()
-  {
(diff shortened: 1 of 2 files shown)


daemon/boardd/queue.hpp
```diff
@@ -2,64 +2,67 @@
...
 #ifndef _QUEUE_H_
 #define _QUEUE_H_
-#include <queue>
-#include <mutex>
-#include <condition_variable>
+#include<condition_variable>
+#include<mutex>
+#include<queue>
+#include<stdexcept> //runtime_error
+#include<utility> //forward, move
```
wens, line 9: Please keep the space...

daemon/boardd/queue.hpp
```diff
@@ -2,64 +2,67 @@
...
+		cv_poped_.notify_one();
+		return true;
+	}
+	void Close()
+	{
+		std::unique_lock<std::mutex> lk{mu_};
+		eof_=true;
+		cv_pushed_.notify_all();
+		cv_poped_.notify_all();
+	}
```
wens, line 65: Please don't reindent the whole file. It makes changes less obvious, and doesn't play nice with blame either.
Use whatever code style the original file had.

Fdhvdu, line 65: How do I fix it?

daemon/boardd/queue.hpp
```diff
@@ -2,64 +2,67 @@
...
+		std::unique_lock<std::mutex> lk{mu_};
+		while(true)
+		{
+			if(eof_)
+				throw std::runtime_error{"Queue closed"};
+			if(max_&&qu_.size()>=max_)
+			{
+				cv_poped_.wait(lk); //do you aware spurious wakeup?
+				                    //if not, you may need a semaphore
+				continue;
```
robertabcd, line 38: This continue would do the check, wouldn't it? (L25 on the left side)

daemon/boardd/queue.hpp
```diff
@@ -2,64 +2,67 @@
...
-  std::condition_variable cv_poped_;
-  bool eof_;
-  size_t max_;
+	Queue() noexcept
+		:Queue{}{}
+	Queue(std::size_t max) noexcept
+		:eof_{false},max_{max}{}
+	//probably need enable_if and is_convertible
+	template<class T2>
+	void Push(T2 &&elem)
```
robertabcd, line 27: Why another template here?

daemon/boardd/queue.hpp
```diff
@@ -2,64 +2,67 @@
...
+	}
+	bool Pop(T &elem)
+	{
+		std::unique_lock<std::mutex> lk{mu_};
+		while(qu_.empty())
+		{
+			if(eof_)
+				return false;
+			cv_pushed_.wait(lk);  //again, how about spurious wakeup?
+		}
```
robertabcd, line 53: This while loop would ensure this, wouldn't it?
Please do not just raise problems in comment unless it's something that you can't fix.

daemon/boardd/queue.hpp
robertabcd, line 65: Yeah, please preserve the original coding style.

daemon/boardd/queue.hpp
Fdhvdu, line 38: Yes. You are right.
Even if there is a spurious wakeup, it is safe.

daemon/boardd/queue.hpp
Fdhvdu, line 53: Yes. You are right.
The while loop ensure it.

daemon/boardd/queue.hpp
Fdhvdu, line 27: That is for distinguish l-value and r-value.

PR #11, opened 2016-06-21
C++11 style code

Fdhvdu:
1. delegating constructor
2. noexcept
3. forwarding reference for Queue::Push (with queue::emplace and std::forward)
4. std::move in Queue::Pop


robertabcd:
Oops. I discovered that this file is no longer needed. I deleted it. Sorry for the burden.

PR #12, opened 2016-06-22
replace thread with future and list with vector

Fdhvdu:
1. remove header (list and thread)
2. add header (future and vector)
3. vector is faster than list if you know memory usage
4. future is enough if you don't call thread::native_handle (and faster than thread when constructing)
5. use size_t (or unsigned) to prevent NUM_THREADS greater than int range
6. for (auto &fut : futures)..., these code can remove safely after c++14
7. to make 6. safely, reorder code is necessary (make futures under server)

### DIFF
--- a/daemon/boardd/server.cpp
+++ b/daemon/boardd/server.cpp
@@ -171,7 +171,7 @@ void start_server(const char *host, unsigned short port) {
     futures.reserve(NUM_THREADS);
     for (std::size_t i = 0; i < NUM_THREADS; ++i) {
 	futures.emplace_back(
-		std::async(std::launch::async,ServiceThread,std::ref(io_service)));
+		std::async(std::launch::async, ServiceThread, std::ref(io_service)));
     }
     for (auto &fut : futures) {
 	fut.get();
--- a/daemon/boardd/server.cpp
+++ b/daemon/boardd/server.cpp
@@ -171,7 +171,8 @@ void start_server(const char *host, unsigned short port) {
     futures.reserve(NUM_THREADS);
     for (std::size_t i = 0; i < NUM_THREADS; ++i) {
 	futures.emplace_back(
-		std::async(std::launch::async, ServiceThread, std::ref(io_service)));
+		std::async(std::launch::async,
+		           ServiceThread,std::ref(io_service)));
     }
     for (auto &fut : futures) {	//TODO remove the for loop after C++14 (it is redundant)
 	fut.get();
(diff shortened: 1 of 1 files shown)


daemon/boardd/server.cpp
```diff
@@ -165,14 +165,15 @@ void start_server(const char *host, unsigned short port) {
...
     asio::io_service io_service;
     Server server(io_service, host, port);
-    for (int i = 0; i < NUM_THREADS; ++i) {
-	threads.push_back(std::move(
-		std::thread(std::bind(&ServiceThread, std::ref(io_service)))));
+    std::vector<std::future> futures;
+    futures.reserve(NUM_THREADS);
+    for (std::size_t i = 0; i < NUM_THREADS; ++i) {
+	futures.emplace_back(
+		std::async(std::launch::async,ServiceThread,std::ref(io_service)));
```
robertabcd: Please preserve the style. Keep it under 80 chars/line, and one space after comma.

daemon/boardd/server.cpp
```diff
@@ -165,14 +165,15 @@ void start_server(const char *host, unsigned short port) {
...
     }
-    for (auto &thr : threads) {
-	thr.join();
+    for (auto &fut : futures) {
```
robertabcd: Do you have any pointers to why this can be removed after C++14? Also, please add a TODO here as well.

daemon/boardd/server.cpp
robertabcd: I digged into documentation about [std::async](http://en.cppreference.com/w/cpp/thread/async). It doesn't guaranteed to create the same number of kernel threads.

> The template function async runs the function f asynchronously (potentially in a separate thread which may be part of a thread pool) and returns a std::future that will eventually hold the result of that function call.

We usually set the number of threads greater than the number of cores we have, due to the application is usually IO-bounded. We need to have control on how many kernel threads the application creates.

Fdhvdu: That is why I use launch policy (to force it to create an "new" thread).
If you don't specify the launch policy, it may execute in "deferred" (in same thread) or "async" (in another thread).
And I specify the launch policy to async.

daemon/boardd/server.cpp
Fdhvdu: http://en.cppreference.com/w/cpp/thread/future/~future

daemon/boardd/server.cpp
Fdhvdu: Do I have to count tabs?

robertabcd:
Let me close this. That file has been removed.

PR #22, opened 2017-09-03
update README and docs/FAQ

holishing:
update README:
    for 17ab073 , 6461cbb and 2801f83

update docs/FAQ:
    log in with empty password for ssh account: bbs , bbsu

improve format:
    replace some TAB with spaces


robertabcd:
I was asking to amend the pmake changes together, but this is fine too. Thanks for your contributions!
You can push -f to PR branch so there's no need to create a new PR.

PR #24, opened 2017-09-25
Add Welly to origin whitelist

clyang:
Welly now supports Websocket connection. It would be great if it can use its own application identity. It might also be helpful for PTT admin to differentiate WSS traffic by clients.


robertabcd:
I have a plan to make this configurable, so the source code can be separated from operations. I will make this open until I've done this.

hungte:
@robertabcd it's one year after, so do you have any plan to move on? Or should we just merge, or close this PR?

robertabcd:
Origin check is now detached from the source code into nginx config as de63cdf1cff06b317a1e3597423c1dd8a3279fc6.

PR #26, opened 2017-09-30
correct symlink of mail for TICKET

holishing:
For solving the issue that user will see 
```
◆ 此封信無內容。                                              [按任意鍵繼續]
```
when they read their mail for getting money by TICKET(全站樂透).
also correct weird symbolic link referring to. (`/home/bbs/home/*/{userid}/etc/ticket`)

### DIFF
--- a/util/openticket.c
+++ b/util/openticket.c
@@ -252,7 +252,7 @@ int main()
 		strcpy(mymail.owner, BBSNAME);
 		sprintf(mymail.title, "[%s] ¤¤¼úÅo! $ %d", Cdatelite(&now), money * num);
 		unlink(genbuf);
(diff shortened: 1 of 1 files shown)


robertabcd:
I think one issue is that file will change over time.

hungte:
Right. I remember one problem we've seen is that this caused people able to make changes to the global ticket file, for example delete records. So we'll probably not fixing it unless that is checked and solved.

hungte (changes requested):
We can't approve this without complete testing how it impacts users when someone deletes files in his home folder.

hungte:
No updates for  10 months. Close inactive PR.

PR #27, opened 2017-10-11
Add reply counter

mlouielu:
Idea from: [[建議] 搜尋「爭議」而不是「推文」數](https://www.ptt.cc/bbs/PttCurrent/M.1487635702.A.29F.html)

Using `fileheader_t` last three pads, split to `unsigned short int reply_counter` and `char pad`. Then it can record up to 65500 reply counts.

I'm not sure about the affect to change the `pad3` in `fileheader_t`, the comment of `filemode` says ` /* must be last field @ boards.c */`.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -226,7 +226,7 @@
 
 #define MAX_MODES	(127)
 #define MAX_RECOMMENDS  (100)
-#define MAX_REPLY_COUNTS  (65550)
+#define MAX_REPLY_COUNTS  (65500)
 
 #define STR_CURSOR      ">"
 #define STR_CURSOR2     "¡´"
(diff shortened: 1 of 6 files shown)


robertabcd:
I think this information can be retrieved from commentd.

mlouielu:
@robertabcd Can commentd calculate each posted reply, and used it in search?

holishing:
Maybe `commentd.py` can really figure out how many comment are replyed in each post ...
and maybe it is more to be improved so that it can be used in search?
And I guess some data are stored in `{BBSHOME}/pttbbs/daemon/commentd/db_comments.db/` ?
(Not very sure)

hungte:
We don't want to hack fileheader_t in a dangerous way, especially that it can be (and should be) done by commentd.

PR #34, opened 2017-12-11
Try to integrate googletest as unit-test.

chhsiao1981:
Try to integrate googletest as unit-test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
-SUBDIR=	common mbbsd util
+SUBDIR=	common mbbsd util tests
 
 .include <bsd.subdir.mk>
 
 .ORDER: all-common all-mbbsd
 .ORDER: all-common all-util
+.ORDER: all-mbbsd all-tests
+.ORDER: all-util all-tests
--- a/docs/unittest.md
+++ b/docs/unittest.md
@@ -1,0 +1,22 @@
+Unit-Test
+===========
+
+這個 unit-test 是 based on googletest. 在 ubuntu 裡使用 libgtest-dev.
+
+Setup (在 ubuntu 16.04 裡)
+-----
+1. `apt-get install -y libgtest-dev valgrind`
+
+2. 在 pttbbs 做 `make GTEST_DIR=/usr/src/gtest`
+
+3. mkdir -p data_test
+
+4. 執行相對應的 test (ex: tests/test_mbbsd/test_dummy)
+
+5. 可執行 `scripts/test_all.sh` 執行所有的 tests
+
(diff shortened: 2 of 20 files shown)


wens:
I don't like submodules. And I don't like vendoring.

What's wrong with using packages provided by the OS?

chhsiao1981:
I am not very familiar with googletest, but the official website seems to recommend compiling googletest as part of source code in the test part.

The other possibility is to not use googletest and use other framework instead.

https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal

chhsiao1981:
Since pttbbs is written in pure c,

it's possible that we just switch to check (looks like the project is still update-to-date):

https://libcheck.github.io/check/index.html

robertabcd:
I hope to see real unit tests written before choosing the framework.
Google test is usually in `libgtest-dev` if you are on Debian/Ubuntu. I prefer this over submodules. Also, try not to use two build systems.

chhsiao1981:
1. I modified the include for googletest as from libgtest-dev. There should be no vendoring issue now～
    (I try to use cmake -DGTEST_DIR=/usr/src/googletest to solve the gtest-compiling issue～
     Hopefully this approach can be cross-platform adaptable～)

2. About the make / cmake thing, I am aware of this issue,
    but the purpose of this pr is to have some unit-test available for the project first,
   and then we can start to do some refactoring more stably.

   We can always improve the make / cmake thing in the future.

3. In terms of "real unit-test", I'll apply this unit-test to the refactoring soon.


https://github.com/chhsiao1981/pttbbs/tree/hsiao.googletest/tests

chhsiao1981:
Intended to use Makefile (pmake) instead of cmake.

For now:
1. Need to separate mbbsd/Makefile with mbbsd/mbbsdvars.mk
2. Need to ignore many .o in mbbsd due to tangling with mbbsd.o, and there is "main" in mbbsd.o
     conflicting "main" in test.

chhsiao1981:
hi @robertabcd 

I revised this PR and provided some basic examples as the templates for the unittest.
(tests/test_common/test_sys/test_string.cc, 
 tests/test_mbbsd/test_bbs.cc)

I added tests/test_mbbsd/mbbsd_mock.c as mock for mbbsd.c because there is main in mbbsd.c
The only difference between mbbsd_mock.c and mbbsd.c is that main becomes mock_main.

Hopefully the revision satisfies the PR and others can contribute more on unittest.

chhsiao1981:
I'll close this for now and revise the pr.

PR #36, opened 2017-12-24
[WIP, DO NOT MERGE!] Hsiao.googletest2

chhsiao1981:
This is to reflect the #34 by integrating CMakeLists.txt in mbbsd / common / util as well.

Tested in Mac (with ncurses installed with homebrew as static lib)
and Ubuntu 16.04 (by export CC=/usr/bin/clang and export CXX=/usr/bin/clang++)

The compiling / linking flags are the same as the original Makefile in these 2 platforms.

This PR is intended to used as discussion. I'll have the corresponding modification on #34 
after the discussion is finalized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 tags
 var.h
 /pttbbs.conf
+/pttbbs.cmake.conf
 /local.h
 common/sys/big5.c
 mbbsd/mbbsd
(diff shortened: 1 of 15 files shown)


chhsiao1981:
Close this PR to prevent accidentally merging.

PR #43, opened 2018-04-06
Specify that Python environment in some utilities are Python 2.x.

holishing:
Although most development environment assume `python` command as Python 2.x environment.

However, in some platform, especially some distribution with rolling release,
default `python` command is specified with Python 3.x environment.
but python utilities here are mostly designed for Python 2.x environment

So we may not assume all `python` command as Python 2.x environment.
unless some modules are really compatible with both Python 2.x and Python 3.x.


holishing:
for #44 , closed.

PR #44, opened 2018-04-27
big5: let python script both compatible with python2 and python3

holishing:
1. `print()` : change usage form `print "string"` to `print ("string")`
2. `print()` : change usage for some string not end with `\n`, and add `from __future__ import print_function` to let it compatible with python2.
3. let byte-like format can also be striped.
4. let `range()` format be converted to `list()` format

ps. 
there's my way to test if generated `big5.c` is still same:
```
python2 big5_gen.py > big5.c
curl -L https://raw.githubusercontent.com/clamtestbbs/pttbbs/r2.pyfix/common/sys/big5_gen.py > big5_gen_test.py
python2 big5_gen_test.py > big5_test2.c
python3 big5_gen_test.py > big5_test3.c
git diff --no-index big5.c big5_test2.c
git diff --no-index big5.c big5_test3.c
```

### DIFF
--- a/common/sys/big5_gen.py
+++ b/common/sys/big5_gen.py
@@ -18,17 +18,17 @@ b2u = [line.strip().split(b' ')
        if line.strip().startswith(b'0x')]
 b2u = dict((int(b, 0), int(u, 0)) for (b, u) in b2u)
 
-print ("""#include <stdint.h>
+print("""#include <stdint.h>
 extern const uint16_t b2u_table[];
 extern const uint16_t u2b_table[];
 extern const uint8_t b2u_ambiguous_width[];
 """)
-print ("const uint16_t b2u_table[0x10000] = {")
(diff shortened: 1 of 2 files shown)


common/sys/big5_gen.py
```diff
@@ -12,37 +13,37 @@
...
 b2u = B2U_FILE.readlines()
-b2u = [line.strip().split(' ')
+b2u = [line.strip().split(b' ')
        for line in b2u
-       if line.strip().startswith('0x')]
+       if line.strip().startswith(b'0x')]
 b2u = dict((int(b, 0), int(u, 0)) for (b, u) in b2u)
 
-print """#include <stdint.h>
+print ("""#include <stdint.h>
```
robertabcd: Can you remove the space between 'print' and '(' ? Looks more like a function call.

PR #42, opened 2018-03-29
Support migration

chhsiao1981:
This is the follow-up of #40.

Adding support with migrate-from-file-to-db and migrate-from-db-to-file.

I'll rewrite edit_post as demo for the new storage framework.

PR #45, opened 2018-05-13
[WIP-DONT-MERGE] Hsiao.pttdb vedit3

chhsiao1981:
This is the follow-up of #42 

The editing part is still work-in-progress.
but it demonstrates the feasibility of the storage-design.

The corresponding docker-image is in the following link:
(docker run chhsiao1981/pttbbs :hsiao.pttdb-vedit3 and then telnet localhost 3000)

https://hub.docker.com/r/chhsiao1981/pttbbs/

The features of the editor (It's used in "重新編輯" (Ctrl-E) for now) is as follow:
1. able to migrate from file-based storage to the (mongo) db-based storage and
    separate main-content / comment / comment-reply in the db-based.

2. Based on the previous feature, we are able to have users with different operations in different content-type.
     ex: Users can only do move-up / move-down in the comment
           (and delete-comment / add comment-reply in the future)

3. Use multithreads to prefetch buffer. Each forked process spawns 1 buffer-thread, trying to prefetch data into buffer.


robertabcd:
How am I supposed to review this?

chhsiao1981:
hi @robertabcd,

Thanks for the reply～

I think for now we can focus on the pttdb part.

You can (assuming that the port 3000 is not used yet):
1. docker run --rm -itd -p 3000:3000 -p 3001:22 --name pttbbs chhsiao1981/pttbbs:hsiao.pttdb-vedit3
2. telnet localhost 3000 and register SYSOP.
3. post 1 article (The original vedit2)
4. do comments on the article.
5. do Ctrl-E to edit the article (it starts the vedit3_wrapper in vedit3.c)
6. docker exec -it pttbbs /bin/bash 
     6.1 mongo localhost/post
     6.2 show collections (and you should be able to see the corresponding collections)
     6.3 db.main.find()
     6.4  db.main_content.find()
     6.5 db.comment.find() (and you should be able to see the corresponding data)
7. try to have the article with 120 more lines, and you should be able to see tmp dir in home/S/SYSOP/storage/
8. When you move to comment-area, you will find it that the allowed operations is only move-up / move-down (which is within expectation for now)
9. Ctrl-X for save-and-exit.
10. read the article and see that the article is updated.

Sometimes I did encounter some timeout in the lock-waiting and resulted in unexpected results though
(catched with error-code, I did not respond the error-code correctly for now～)
The lock is in the pttui-part.
I am trying to reduce the probability of the timeout.

Thanks!

robertabcd:
You didn't answer my question.

chhsiao1981:
hi @robertabcd 

Basically this PR is kind of like a demo that the performance concern about db
can be worked-around with the appropriate buffer-prefetch mechanism.

I'll try my best to prepare a design-document (in Chinese)～

robertabcd:
> Basically this PR is kind of like a demo that the performance concern about db
can be worked-around with the appropriate buffer-prefetch mechanism.

What performance concern are you addressing particularly? And how do you reach this conclusion?

chhsiao1981:
hi @robertabcd ,

Originally there is a concern that compared to file-based storage (storage of current setup)
There may be performance concern about db-based storage.

This PR is kind of like a demo that we can use multithreads and do buffer-prefetching
to work-around this issue～

robertabcd:
Why is your "buffer-prefetching" work-around the issue?  I still don't understand.  You are fetching data for editing, but not for reading.  And did you test when there are multiple readers?

chhsiao1981:
hi @robertabcd 

The design is that:
1. The article is decomposed as main-content (本文). comment(s) (推文). comment-reply(s) (推文的回文)

2. main-content / comment-reply is represented as meta-data, and with the corresponding content-blocks
     as the raw-data.
     (本文和推文的回文. 都是有一個 header 的 struct 做為這個文章的 meta-data. 會指到現在 available 的相對應的 content-id (一個 content-id 會有多個 fixed-size 的 blocks))

3. The editing is similar to what it is now: Instead of directly modifying the original content, we store the modified data in the local-temporary-files.

4. When doing editing, we obtain the "newest-comment-until-now", and construct the whole file-info meta-data (pttdb_file_info.h), as the reference of the whole file.

5. Buffer-prefetch intends to solve the consistency issue between db / file / screen.

6. For the updated-content (main-content / comment-reply): we update the content-blocks first (with new content-id) to ensure the correctness of the data-in-db.
     and then update main-content / comment-reply (atomic-update) to point to the new content-id.
     (對於本文和回文的 update: 先對於 content update 回 db (新的 content-id). 確定 content 在 db 裡是 ok 以後. 再將新的 content-id update 到本文 / 回文的 header)

7. Updated-comments (已被 xxx 刪除: [原因]) are stored in the local-temporary-files as well, and directly update back to comments (atomic-update)

robertabcd:
Please don't repeat your design. I already have some idea from many PRs your sent.
You said earlier:
> There may be performance concern about db-based storage.

But now you are saying:
> 5. Buffer-prefetch intends to solve the consistency issue between db / file / screen.

My question to the former one is: why the "buffer-prefetching" works around this issue?
And to the latter one is: why "buffer-prefetching" solves consistency issue?
I don't feel like we fully understand each other.

chhsiao1981:
hi @robertabcd 

Thanks for the reply～
I am sorry that I did not clearly what I intended to say about "buffer-prefeteching" in the last reply.

Given that the cursor is at the ith file-line (and know the corresponding line-info)
and know that the top-window-line is at the jth file-line (and the corresponding line-info)

Buffer-prefeteching utilizes multithread to do the following things in the background:
1. Buffer is constructed with doubly-linked-list, with knowing head / tail.
2. try to load n-lines before the top-window-line and n-lines after the top-window-line (n = 250 for now)
    It's with the hope that the buffer is large enough to cover the whole window and the possible following 1~2 operations (move-up / move-down / move-pgdn / move-pgup)
 
3. if top-window-line is close to top-buffer-line (distance < 100 for now), try to load 250 more lines before the top-buffer-line. if top-window-line is close to tail-buffer-line (distance < 100) for now, try to load 250 more lines after tail-buffer-line.

4. if the buffer size reaches N (N = 600 for now). do the shrinking.
     (I am thinking to increase the N to 800 or 1000. I need to do some experiments to determine the N though～)

Currently there is some issue with buffer-prefeteching about syncing between db / file / screen when deleting lines.

chhsiao1981:
Due to the huge difference between the main repo and this branch,
looks like it's better that I'll just have a separated repo before it is
well tested.

I'll close the related PRs for now～

PR #47, opened 2018-05-24
Add googletest for unit-test framework 

chhsiao1981:
This PR is based on #46 

Several test-files as the examples of  writing unit-test.
Also add the dockerfile-for-googletest to easily do the unit-test.

PR #49, opened 2018-05-31
Add multithread.

chhsiao1981:
This is the follow-up of #47 

Dockerfile: chhsiao1981/pttbbs:hsiao.pttui-thread

Currently user-interaction and buffer-retrieval are in the same thread.
We do not feel the performance issue because of relatively high-performance of
the single-file-based storage-operations.
(main-content and comments are in the same file, and using mmap to boost the performance)

Main-content and comments may be separated in the future, which may introduce lower-performance when doing buffer-retrieval.

This PR intends to initiate incorporating 1 more thread specifically for buffer-retrieval.

More information can be reached in docs/pttui_thread.md


chhsiao1981:
Due to the huge difference between the main repo and this branch,
looks like it's better that I'll just have a separated repo before it is
well tested.

I'll close the related PRs for now～

PR #55, opened 2018-12-14
sort out and move some functions to common library

holishing:
some functions are placed in some *special* files seems to due to historical reasons.
here is my way to sort out this functions to better places (maybe).

Some issues I'm not sure if it is ok:
* about licenses: `vtuikit.c` licensed in 2-clause BSD License, but most of other functions in mbbsd licensed in GPLv2. (I still moved some functions because some related annotations mentioned that)
* some new places the functions moved to I'm not sure if is properly

so some suggestions or any other ideas needed, thanks!


holishing:
related build logs for reference:
https://cloud.docker.com/u/bbsdocker/repository/registry-1.docker.io/bbsdocker/imageptt/hub-builddetail/bjeka3e6xlnw38rfebj9wqn

wens:
The common libraries should not be using any global variables, especially those defined only for mbbsd.
This includes variables like cuser, currmode, etc..

holishing:
Thanks for suggestion, I will consider how to make better revision and send PR again.

PR #67, opened 2019-10-14
Fix a Chinese grammatical error

johannjhang:


kcwu:
I guess it was 人次 because one account can login more than once.

johannjhang:
> I guess it was 人次 because one account can login more than once.

Usually the definition of [人次](https://www.moedict.tw/%E4%BA%BA%E6%AC%A1) is within a period of time, how many people have come and gone, or passed by, cumulatively. While "同時在線" is a number that represents the number of people in a specific area or a concept of a field, right now, which is related to neither periods of time nor accumulation so "次" doesn't make sense here.

So I guess we'll always say "同時在線人數" but never "同時在線人次" even though one person can be recognized as more than one by the system if he/she establishes more than one connection at once.

PR #68, opened 2019-11-02
Add a new feature to mark the posts that need further check

sitos:


common/bbs/search.c
```diff
@@ -38,10 +38,13 @@ match_fileheader_predicate(const fileheader_t *fh, void *arg)
 	return DBCS_strcasestr(fh->title, keyword) == NULL;
     else if (sr_mode & RS_TITLE)
 	return strcasecmp(subject(fh->title), keyword) == 0;
-    else if (sr_mode & RS_RECOMMEND)
-	return pred->recommend > 0 ?
+    else if (sr_mode & RS_RECOMMEND) {
+        if (fh->filemode & FILE_SUSPICIOUS)
+            return 0;
```
robertabcd: Can this entry be treated as `fh->recommend` being `0`, not silently pretending it is not matching?

include/pttstruct.h
```diff
@@ -284,23 +284,24 @@ typedef struct fileheader_t { /* 128 bytes */
 	} refer;
     }	    multi;		    /* rocker: if bit32 on ==> reference */
     /* XXX dirty, split into flag and money if money of each file is less than 16bit? */
-    unsigned char   filemode;        /* must be last field @ boards.c */
-    char    pad3[3];
+    uint16_t   filemode;        /* must be last field @ boards.c */
```
robertabcd, line 287: Note this is assuming little endian. @wens

mbbsd/bbs.c
```diff
@@ -2380,6 +2396,11 @@ read_post(int ent, fileheader_t * fhdr, const char *direct)
     } while (0);
 #endif
 
+    if (fhdr->filemode & FILE_SUSPICIOUS)
+    {
+        more(FN_FILE_SUSPICIOUS_DESC, YEA);
```
robertabcd, line 2403: This will probably break many mobile apps.

mbbsd/bbs.c
```diff
@@ -3690,6 +3711,32 @@ mark_post(int ent, fileheader_t * fhdr, const char *direct)
     return change_post_mode(ent, fhdr, direct, FILE_MARKED);
 }
 
+static int
+mark_suspicious(int ent, fileheader_t * fhdr, const char *direct)
+{
```
robertabcd, line 3718: People may accidentally hit `M` when they try to `m` the post if they have their caps lock on.

I think we should just implement this in `{ 1, manage_post }, // Ctrl('E')`.

mbbsd/bbs.c
```diff
@@ -3690,6 +3711,32 @@ mark_post(int ent, fileheader_t * fhdr, const char *direct)
...
+        aidu2aidc(aidc, aidu);
+
+        if (fhdr->filemode & FILE_SUSPICIOUS)
+            snprintf(buf, sizeof(buf),
+                     "%s �N #%s (%s) �Ѱ��ݬd��", cuser.userid, aidc, bp->brdname);
+        else
+            snprintf(buf, sizeof(buf),
+                     "%s �N #%s (%s) �аO�ݬd��", cuser.userid, aidc, bp->brdname);
+
+        post_msg(BN_SECURITY, buf, "�Ъ`�N�аO���X�k��", "[���ҼаO]");
```
robertabcd, line 3733: The user needs to pass the permission check (`MODE_BOARD`) before they can do this. And this is done in `change_post_mode`. 
This will cause erroneous logging for normal users.

include/pttstruct.h
wens: Yes we can assume little endian. There aren't many big endian Linux distro ports anymore, at least not for modern hardware. I doubt the code base was ever tested on big endian either. It might be full of bugs lol.

kcwu: Don't forget somebody have to clear the pad to zero before deploy this code.

common/bbs/search.c
sitos: Thanks. I changed the implementation to treat the file as ```fh->recommend``` being ```0```.

mbbsd/bbs.c
sitos: Thanks. ```mark_suspicious``` has been moved into ```manage_post```.

mbbsd/bbs.c
sitos: Because ```mark_suspicious``` was moved into ```manage_post``` and the permission check is done at the beginning of ```manage_post```, I think there is no need to check the permission in ```mark_suspicious``` redundantly.

mbbsd/bbs.c
```diff
@@ -129,7 +129,7 @@ modify_dir_lite(
     if (disable_modes)
         fhdr.filemode &= ~disable_modes;
     if (title && *title)
-	strlcpy(fhdr.title, title, sizeof(fhdr.title));
+        strlcpy(fhdr.title, title, sizeof(fhdr.title));
```
wens, line 132: Unrelated change.

mbbsd/bbs.c
```diff
@@ -3690,6 +3711,32 @@ mark_post(int ent, fileheader_t * fhdr, const char *direct)
...
+    aidu_t aidu = 0;
+    aidu = fn2aidu((char *)fhdr->filename);
+    if (aidu > 0) {
+        char aidc[10];
+        char buf[256];
+        aidu2aidc(aidc, aidu);
+
+        if (fhdr->filemode & FILE_SUSPICIOUS)
+            snprintf(buf, sizeof(buf),
+                     "%s �N #%s (%s) �Ѱ��ݬd��", cuser.userid, aidc, bp->brdname);
```
wens: `currboard` can be used instead of `getbcache(currbid)` if all you need is the board name.

mbbsd/bbs.c
robertabcd: This is just FYI. Mark as resolved.

robertabcd:
LGTM. Please also check comments from wens.

sitos:
Hi, I will revise the codes for wens' comments, and I will fix the display issue that changes the color of the next line (need ANSI_RESET for file_suspicious near line 878).

wens:
For the record this is already running on PTT.
However we should wait until there is a utility to clear the newly used fields before merging this.
I have a semi-working version that only deals with boards and not man/boards (no recursion).

hungte:
It has been three years and given "this is already running on PTT" I tend to close the PR. We may revisit when the "semi-work" is done.

sitos:
Thanks for closing this. :)

PR #70, opened 2019-11-18
Adjust Ubuntu build settings

Truth1987:
As title. 
Adjust Ubuntu build settings and integrate Wiki content.


wens:
The sample config file mentions that `SHMALIGNSIZE` is only needed when `HUGETLB` is used.
I believe that is still the case. So yes, standard builds on Linux and FreeBSD don't need `SHMALIGNSIZE` set, but it would be better to describe under what conditions it is required.

PR #73, opened 2020-02-18
Email verification at account registration

robertabcd:


mbbsd/register.c
```diff
@@ -563,6 +599,108 @@ do_register_captcha()
...
+		pressanykey();
+		continue;
+
+	    default:
+		return err;
+	}
+    } while (err != REGISTER_OK);
+
+    // Send and check regcode.
+    char inregcode[14] = {0}, regcode[50];
```
wens: The number 13 is scattered everywhere. We should make it a macro.

mbbsd/register.c
```diff
@@ -563,6 +599,108 @@ do_register_captcha()
...
+	outs("�{�ҽX�`�@����Q�T�X, �S���ťզr��.\n");
+	outs("�Y�L�[�S����{�ҽX�i��J x ���s�H�e.\n");
+
+	getdata(20, 0, "�п�J�{�ҽX�G",
+		inregcode, sizeof(inregcode), DOECHO);
+	if (strcasecmp(inregcode, "x") == 0) {
+	    send_code = true;
+	    continue;
+	}
+	if (strcasecmp(inregcode, regcode) != 0) {
```
wens, line 697: Maybe also handle empty string separately? Like not count it as an attempt?

mbbsd/register.c
```diff
@@ -280,23 +304,35 @@ delregcodefile(void)
...
-	 * subjects (BBSNAME) correctly, so we'd like to use
-	 * BBSENAME here to prevent subject being messed up.
-	 * And please keep BBSENAME short or it may be truncated
-	 * by evil mail servers.
-	 */
-	snprintf(buf, sizeof(buf),
-		 " " BBSENAME " - [ %s ]", makeregcode(genbuf));
-
-	bsmtp("etc/registermail", buf, muser->email, "non-exist");
+	char regcode[256];
```
wens: Use a macro. No need for such a large buffer.

mbbsd/register.c
```diff
@@ -1078,12 +1078,78 @@ create_regform_request()
...
+#define REGISTER_ERR_EMAILDB (-2)
+#define REGISTER_ERR_TOO_MANY_ACCOUNTS (-3)
+#define REGISTER_ERR_CANCEL (-4)
+
+static int
+register_email_input(const char *userid, char *email)
+{
+    while (1) {
+	email[0] = 0;
+	getfield(15, "����{�ҥ�", REGNOTES_ROOT "email", "E-Mail Address", email, 50);
```
wens: For future reference, need to clean up hardcoded 50 length.

wens (changes requested):

mbbsd/register.c
robertabcd: Made a constant. See later commits.

mbbsd/register.c
robertabcd: In the rest of registration flow, empty input is treated as an error, or a failed attempt.

mbbsd/register.c
robertabcd: Done.

mbbsd/register.c
robertabcd: Made a macro for this.

mbbsd/register.c
wens: Ok. Consistency is good.

PR #76, opened 2020-02-23
Various improvements to registration flow

robertabcd:
See commit messages.

### DIFF
--- a/mbbsd/register.c
+++ b/mbbsd/register.c
@@ -633,10 +633,11 @@ new_register_email_verify(char *email)
 	    case REGISTER_OK:
 		if (strcasecmp(email, "x") != 0)
 		    break;
-		// Use input is "x".
+		// User input is "x".
 		err = REGISTER_ERR_INVALID_EMAIL;
 		// fallthrough
 	    case REGISTER_ERR_INVALID_EMAIL:
+	    case REGISTER_ERR_CANCEL:
 		move(15, 0); clrtobot();
 		move(17, 0);
 		outs("指定的 E-Mail 不正確。可能你輸入的是免費的Email，\n");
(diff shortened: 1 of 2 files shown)


mbbsd/register.c
```diff
@@ -597,6 +597,17 @@ do_register_captcha()
...
+check_register_connection_security()
+{
+#ifdef REQUIRE_SECURE_CONN_TO_REGISTER
+    if (mbbsd_is_secure_connection())
+	return;
+    vmsg("�ШϥΦw���s�u��U�b��!");
+    exit(1);
+#endif
+}
+
```
wens, line 618: The function is only used once, and does not help with readability. Just inline it.

mbbsd/register.c
```diff
@@ -916,27 +935,24 @@ new_register(void)
...
+    // Whether to limit login to secure connection only.
+    if (mbbsd_is_secure_connection()) {
+	// Screen full.
+	y = 17;
+	move(y, 0); clrtobot();
+	outs("[ �s�u�]�w ]");
+
+	y++;
+	if (query_yn(y, "�z�O�_�n����b���ȯ�ϥΦw���s�u�n�J? [y/n]:"))
+	    newuser.uflag |= UF_SECURE_LOGIN;
```
wens, line 948: I thought you wanted to make this the default? lol

mbbsd/register.c
robertabcd: OK. Inlined.

mbbsd/register.c
robertabcd: Yeah, but later I thought that user might not know what a secure connection even is... So I made them to be responsible for their own choice instead.

wens (approved):

Issue #1, opened 2015-08-04
pmore hang


kcwu:
How to reproduce
  use this https://github.com/kcwu/pttbbs-fuzzy

```
cd mbbsd
make pmore
echo '0000000: 0c 23 00 2c 66 31 23 0a 30 ab ab ab ab ab ab     .#.,f1#.0......' | xxd -r > input
./pmore input
```

detected by afl-fuzz

wens:
@kcwu @hungte Could you guys review this?

holishing:
I wonder if this issue is actually fixed by @IepIweidieng in #77

IepIweidieng:
The hanging caused by the specific case mentioned is fixed in #77.

There may still be unknown cases which lead to infinite loops like this.

However, I believe that it would be better to close this issue than to keep the issue open until such cases are actually found.

wens:
Thanks.

PR #77, opened 2020-03-19
Fix pmore hanging by making pmore handle `'\0'`

IepIweidieng:
This fixes issue #1.

This command makes pmore hang:
```
0c  23  00  2c  66  31  23  0a  30  ab  ab  ab  ab  ab  ab
```

The command in the form of mix of caret notation and C escape sequences:
```
^L   #  \0   ,   f   1   #  \n   0\xab\xab\xab\xab\xab\xab
```

The command in plaintext with escapes:
```
^L#\0,f1#
0\xab\xab\xab\xab\xab\xab
```

Where:
```
^L#\0,f1#
```
is a pmore command which sets up a interactive choice with one option without the prompt.
It literally means to go to the 1st frame when the user send `'\0'` (can be sent with `Ctrl-Shift-2` on PCMan), although `'\0'` is an invalid choice key.

```
0\xab\xab\xab\xab\xab\xab
```
is the content of the 1st frame.
Actually, put any non-empty string here will make pmore hang.

However, there is a logic error in `mf_movieOptionHandler()`:
```c
                // handle selection
                if (c == key ||
                        (key == MOVIE_KEY_ANY && c != 0))
                {
                    // hotkey pressed
                    selected = 1;
                    isel = maxsel;
                }
```
When `^L#\0,f1#` is processed, the variable `key` is `'\0'`, and the variable `c` is its initial value, which is also `0`.

This makes the program think that the hotkey `'\0'` is pressed even if the user presses nothing, and therefore no user inputs will be processed.

Then the command `f1` will be executed in `mf_movieExecuteOffsetCmd()`, making the pmore interpreter go to the 1st frame, where `^L#\0,f1#` resides.

The pmore interpreter will then enter a infinity loop which repeatly executes `^L#\0,f1#` and skips user inputs due to the logic error aforementioned, which makes the interpreter ignore user inputs and hang forever.

There are 2 ways to fix the issue:

1. Modify the condition to always check `c` against `0`:
```c
                if ((c == key || key == MOVIE_KEY_ANY) && c != 0)
```
   This is the simple way but is only a workaround.

2. Introduce another special `vget()` value which `vget()` never returns, to distinguish _nothing_ from `'\0'`.
   This also makes pmore correctly handle `'\0'` as a normal key (but it is still an invalid choice key).

The 2nd way is chosen since it make the intention of the code clearer.

I reproduced the issue and tested the fix by modifying the function `main()` of `mbbsd`:
```diff
--- a/mbbsd/mbbsd.c
+++ b/mbbsd/mbbsd.c
@@ -1770,6 +1770,26 @@ bool parse_argv(int argc, char *argv[], struct ProgramOption *option)
 int
 main(int argc, char *argv[], char *envp[])
 {
+    init();
+    init_tty();
+    do_term_init(TermMode_TTY, 80, 24);
+
+    set_converting_type(CONV_UTF8);
+
+    if (argc < 2)
+    {
+        unsigned char data[] = {0x0c, 0x23, 0x00, 0x2c, 0x66, 0x31, 0x23, 0x0a, 0x30};
+        pmore2_inmemory(data, sizeof(data), 1, NULL, NULL, NULL, NULL);
+    }
+    else
+    {
+        pmore2(argv[1], 1, NULL, NULL, NULL, NULL);
+    }
+
+    vmsg("complete. enter to exit.");
+
+    return 0;
+
     bool oklogin = false;
     struct ProgramOption *option;

```

Then enter the root directory of the project and run:
```shell
$ bmake
$ echo '0000000: 0c 23 00 2c 66 31 23 0a 30 ab ab ab ab ab ab     .#.,f1#.0......' | xxd -r > input
$ (sleep 10; pkill mbbsd)& mbbsd/mbbsd input
```

### DIFF
--- a/mbbsd/pmore.c
+++ b/mbbsd/pmore.c
@@ -741,6 +741,7 @@ MFFPROTO int mf_movieMaskedInput(int c);
 
 // some magic value that your vkey() will never return
 #define MOVIE_KEY_ANY (0x4d464b41)
+#define MOVIE_KEY_NONE (0x414b464d)
 
 // some environments already converted 0x7F to 0x08,
 // but we'd still do it again here for compatibility.
@@ -3261,7 +3262,7 @@ mf_movieWaitKey(struct timeval *ptv, int dorefresh)
 {
     int sel = 0;
     fd_set readfds;
-    int c = 0;
+    int c = MOVIE_KEY_NONE;
 
     if (dorefresh)
         refresh();
@@ -3309,7 +3310,7 @@ mf_movieWaitKey(struct timeval *ptv, int dorefresh)
         syncnow();
 #endif // PMORE_HAVE_SYNCNOW
 
-    return (sel == 0) ? 0 : 1;
+    return (sel == 0) ? MOVIE_KEY_NONE : 1;
 }
 
 // type : 1 = option selection, 0 = normal
@@ -3465,7 +3466,7 @@ mf_movieNamedKey(int c)
         default:
             break;
     }
-    return 0;
+    return MOVIE_KEY_NONE;
 }
 
 MFFPROTO int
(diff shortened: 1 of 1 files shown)


wens:
@kcwu @hungte Could you guys review this?

PR #81, opened 2020-06-17
Support multiple account verification methods

robertabcd:
**A new `verifydb` table served by `regmaild` that stores unique verification key.**

The current `emaildb` table stores `userid` and `email` only. Since `userid` may be reused after it expires, the server determines whether the entry is up-to-date by comparing `email` with the one in passwd file.

As `userec_t` is full, we can't add more fields to support other verification methods. (i.e. no comparing with passwd file) The new `verifydb` table uses `(userid, generation)` to identify each user. `generation` is the `firstlogin` field in `userec_t`.

To support multiple verification at the same, PK is set to `(userid, generation, vmethod)`.

Email is now dual-write to verifydb.

### DIFF
--- a/daemon/regmaild/regmaild.c
+++ b/daemon/regmaild/regmaild.c
@@ -33,8 +33,11 @@
 #include <sys/resource.h>
 #include <sys/wait.h>
 
-#include "pttstruct.h"
 #include "common.h"
+#include "cmbbs.h"
+#include "daemons.h"
+#include "perm.h"
+#include "pttstruct.h"
 
 ///////////////////////////////////////////////////////////////////////
 // EmailDB Common Section
@@ -111,6 +114,45 @@ emaildb_filter_user(const userec_t *u)
     return true;
 }
 
+static bool
+verifydb_prepare_stmt(sqlite3 *db, sqlite3_stmt **stmt)
+{
+    // Note that we lowercase vkey, as we are only importing email.
+    if (sqlite3_prepare(db, "REPLACE INTO verifydb "
+                "(userid, generation, vmethod, vkey, timestamp) VALUES "
+                "(lower(?),?,?,lower(?),?);",
+                -1, stmt, NULL) != SQLITE_OK)
+    {
+        fprintf(stderr, "Error preparing stmt: %s\n", sqlite3_errmsg(db));
+        return false;
+    }
+    return true;
+}
+
+static bool
+verifydb_bind_args(sqlite3 *db, sqlite3_stmt *stmt, const userec_t *u)
+{
+    // vkey (email) will be lower-cased in sql stmt.
+    if (sqlite3_bind_text(stmt, 1, u->userid, -1, SQLITE_STATIC) != SQLITE_OK ||
+        sqlite3_bind_int64(stmt, 2, u->firstlogin) != SQLITE_OK ||
+        sqlite3_bind_int64(stmt, 3, VMETHOD_EMAIL) != SQLITE_OK ||
+        sqlite3_bind_text(stmt, 4, u->email, -1, SQLITE_STATIC) != SQLITE_OK ||
+        sqlite3_bind_int64(stmt, 5, 0) != SQLITE_OK)
+    {
+        fprintf(stderr, "Error binding args: %s\n", sqlite3_errmsg(db));
+        return false;
+    }
+    return true;
+}
+
+static bool
+verifydb_filter_user(const userec_t *u)
+{
+    return is_validuserid(u->userid) &&
+        (u->userlevel & PERM_LOGINOK) &&
+        strchr(u->email, '@') != NULL;
+}
+
 typedef struct {
     bool (*prepare_stmt)(sqlite3 *db, sqlite3_stmt **stmt);
     bool (*bind_args)(sqlite3 *db, sqlite3_stmt *stmt, const userec_t *u);
@@ -123,6 +165,12 @@ static initemaildb_ops emaildb_ops = {
     .filter_user = emaildb_filter_user,
 };
 
+static initemaildb_ops verifydb_ops = {
+    .prepare_stmt = verifydb_prepare_stmt,
+    .bind_args = verifydb_bind_args,
+    .filter_user = verifydb_filter_user,
+};
+
 #define TRANSCATION_PERIOD (4096)
 int main(int argc, char *argv[])
 {
@@ -132,7 +180,24 @@ int main(int argc, char *argv[])
     sqlite3 *Db = NULL;
     sqlite3_stmt *Stmt = NULL, *tranStart = NULL, *tranEnd = NULL;
     const char *fpath = EMAILDB_PATH;
-    const initemaildb_ops *ops = &emaildb_ops;
+    const initemaildb_ops *ops = NULL;
+
+    if (argc != 2 && argc != 3) {
+        fprintf(stderr, "Usage: %s <emaildb|verifydb> [dbfile]\n", argv[0]);
+        return 1;
+    }
+
+    if (!strcmp(argv[1], "emaildb"))
+        ops = &emaildb_ops;
+    else if (!strcmp(argv[1], "verifydb"))
+        ops = &verifydb_ops;
+    else {
+        fprintf(stderr, "Unknown db: %s\n", argv[1]);
+        return 1;
+    }
+
+    if (argc > 2)
+        fpath = argv[2];
 
     // init passwd
     sz = dashs(FN_PASSWD);
@@ -143,9 +208,6 @@ int main(int argc, char *argv[])
         return 1;
     }
     sz /= sizeof(userec_t);
-
-    if (argc > 1)
-        fpath = argv[1];
 
     // init emaildb
     if (regmaildb_open(&Db, fpath) != SQLITE_OK)
@@ -226,7 +288,6 @@ int main(int argc, char *argv[])
 #include <event.h>
 #include <buffer.h>
 #include "bbs.h"
-#include "daemons.h"
 
 int 
 regmaildb_check_email(const char * email, int email_len, const char *myid)
--- a/daemon/regmaild/regmaild.c
+++ b/daemon/regmaild/regmaild.c
@@ -383,7 +383,7 @@ end:
 }
 
 static void
-verifydb_count(const verifydb_req *req, verifydb_count_rep *rep)
+verifydb_count_handler(const verifydb_req *req, verifydb_count_rep *rep)
 {
     sqlite3 *Db = g_Db;
     sqlite3_stmt *Stmt = NULL;
@@ -443,7 +443,7 @@ end:
 }
 
 static int
-verifydb_set(const verifydb_req *req)
+verifydb_set_handler(const verifydb_req *req)
 {
     int ret = VERIFYDB_ERROR;
 
@@ -491,7 +491,7 @@ end:
 }
 
 static void
-verifydb_get(const verifydb_req *req, int fd)
+verifydb_get_handler(const verifydb_req *req, int fd)
 {
     sqlite3 *Db = g_Db;
     sqlite3_stmt *Stmt = NULL;
@@ -963,7 +963,7 @@ client_cb(int fd, short event, void *arg)
                 goto end;
 
             verifydb_count_rep rep = {};
-            verifydb_count(req, &rep);
+            verifydb_count_handler(req, &rep);
             fprintf(stderr, "verifydb count: %s generation [%ld] vmethod [%d] vkey [%s] "
                     "(status: %d, self: %d, other %d)\n",
                     req->userid, req->generation, req->vmethod, req->vkey,
@@ -979,7 +979,7 @@ client_cb(int fd, short event, void *arg)
             if (!validate_verifydb_request(req))
                 goto end;
 
-            int ret = verifydb_set(req);
+            int ret = verifydb_set_handler(req);
             fprintf(stderr, "verifydb set: %s generation [%ld] vmethod [%d] vkey [%s] (status: %d)\n",
                     req->userid, req->generation, req->vmethod, req->vkey, ret);
             if (towrite(fd, &ret, sizeof(ret)) != sizeof(ret))
@@ -993,7 +993,7 @@ client_cb(int fd, short event, void *arg)
             if (!validate_verifydb_request(req))
                 goto end;
 
-            verifydb_get(req, fd);
+            verifydb_get_handler(req, fd);
             break;
         }
 
--- a/daemon/regmaild/regmaild.c
+++ b/daemon/regmaild/regmaild.c
@@ -859,10 +859,13 @@ validate_verifydb_request(const verifydb_req *req)
                 req->operation);
         return false;
     }
-    if (!*req->userid) {
-        fprintf(stderr, "invalid request: op=[%d], empty userid\n",
-                req->operation);
-        return false;
+    if (req->operation == VERIFYDB_REQ_GET ||
+        req->operation == VERIFYDB_REQ_SET) {
+        if (!*req->userid) {
+            fprintf(stderr, "invalid request: op=[%d], empty userid\n",
+                    req->operation);
+            return false;
+        }
     }
     if (req->operation == VERIFYDB_REQ_COUNT ||
         req->operation == VERIFYDB_REQ_SET) {
--- a/daemon/regmaild/regmaild.c
+++ b/daemon/regmaild/regmaild.c
@@ -77,6 +77,52 @@ regmaildb_open(sqlite3 **Db, const char *fpath) {
 
 // standalone initialize builder
 
+static bool
+emaildb_prepare_stmt(sqlite3 *db, sqlite3_stmt **stmt)
+{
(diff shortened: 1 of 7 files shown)


daemon/regmaild/regmaild.c
```diff
@@ -100,52 +147,47 @@ int main(int argc, char *argv[])
...
+            fprintf(stderr, "\ncannot execute statement: %s\n",
+                    sqlite3_errmsg(Db));
             break;
         }
         sqlite3_reset(Stmt);
 
-        valids ++;
+        valids++;
         if (valids % 10 == 0)
-            fprintf(stderr, "%d/%d (valid: %d)\r",
```
wens: Could you split out the error number fixes and style fixes into separate patches when possible?
That would make it easier to read.

daemon/regmaild/regmaild.c
```diff
@@ -73,6 +73,52 @@ regmaildb_open(sqlite3 **Db, const char *fpath) {
...
+}
+
+static int
+emaildb_filter_user(const userec_t *u)
+{
+    if (strlen(u->userid) < 2 || strlen(u->userid) > IDLEN)
+        return 0;
+    if (strlen(u->email) < 5)
+        return 0;
+    return 1;
```
wens: If all you are returning is success and failure, maybe use `bool` as return type explicitly?

daemon/regmaild/regmaild.c
```diff
@@ -73,6 +73,52 @@ regmaildb_open(sqlite3 **Db, const char *fpath) {
...
+}
+
+static int
+emaildb_bind_args(sqlite3 *db, sqlite3_stmt *stmt, const userec_t *u)
+{
+    if (sqlite3_bind_text(stmt, 1, u->userid, -1, SQLITE_STATIC) != SQLITE_OK ||
+        sqlite3_bind_text(stmt, 2, u->email, -1, SQLITE_STATIC) != SQLITE_OK)
+    {
+        fprintf(stderr, "\nError binding args: %s\n", sqlite3_errmsg(db));
```
wens: Get rid of the leading newline?

daemon/regmaild/regmaild.c
```diff
@@ -107,6 +109,44 @@ emaildb_filter_user(const userec_t *u)
...
+    // vkey will be lower-cased in sql stmt.
+    char vkey[EMAILSZ+16];
+    snprintf(vkey, sizeof(vkey), "email:%s", u->email);
+
+    if (sqlite3_bind_text(stmt, 1, u->userid, -1, SQLITE_STATIC) != SQLITE_OK ||
+        sqlite3_bind_int64(stmt, 2, u->firstlogin) != SQLITE_OK ||
+        sqlite3_bind_int64(stmt, 3, 0) != SQLITE_OK ||
+        sqlite3_bind_text(stmt, 4, vkey, -1, SQLITE_TRANSIENT) != SQLITE_OK)
+    {
+        fprintf(stderr, "\nError binding args: %s\n", sqlite3_errmsg(db));
```
wens: Drop the leading newline.

wens:
So this supports multiple options for verification, but not multiple concurrent options, i.e. both email and mobile are verified. Am I right?

robertabcd:
The schema now supports multiple verification methods for an user at the same time. Also dual-write implemented.

Other verification methods will need more refactoring, and will come in separate PRs.

daemon/regmaild/regmaild.c
```diff
@@ -278,12 +272,206 @@ regmaildb_update_email(const char * userid, int userid_len, const char * email,
...
+            fprintf(stderr, "verifydb count: skipped old record: %s "
+                    "(generation: db %ld, current %d)\n",
+                    u.userid, generation, u.firstlogin);
+            continue;
+        }
+
+        if (strcasecmp(userid, req->userid) == 0)
+            rep->count_self++;
+        else
+            rep->count_other++;
```
wens, line 429: This is almost certainly not going to match, because of the generation check above.
What you might want is to place the generation check inside the userid match case.

daemon/regmaild/regmaild.c
```diff
@@ -278,12 +272,206 @@ regmaildb_update_email(const char * userid, int userid_len, const char * email,
...
+    }
+
+    rep->status = VERIFYDB_OK;
+
+end:
+    if (r != SQLITE_OK)
```
wens, line 435: You can use multiple goto targets, one for each failure case.
That way you can get rid of some the if tests below.
This is used extensively in the kernel.

daemon/regmaild/regmaild.c
```diff
@@ -278,12 +272,206 @@ regmaildb_update_email(const char * userid, int userid_len, const char * email,
...
+        fprintf(stderr, "sqlite3: %d %s\n", r, sqlite3_errmsg(Db));
```
wens, line 436: Are the standard sqlite3 error messages specific enough that you can tell which function above caused it?
If not, it might be best to have more detailed error messages directly following the error.

daemon/regmaild/regmaild.c
```diff
@@ -516,87 +704,204 @@ regcheck_ambiguous_id(const char *userid)
 //
 // operation: set,count
 
-static void
-err_request(regmaildb_req *req, int fd, struct event *ev)
+static bool
+check_null_terminated(const char *s, size_t sz)
```
wens, line 812: `return !!memchr(s, '\0', sz);` ?

daemon/regmaild/regmaild.c
```diff
@@ -516,87 +704,204 @@ regcheck_ambiguous_id(const char *userid)
...
+        case REGCHECK_REQ_AMBIGUOUS:
+            // No checks.
+            break;
+
+        default:
+            fprintf(stderr, "unknown request: op=[%d]\n", req->operation);
+            return false;
+    }
+    return true;
+}
```
wens, line 847: This change and other cleanup of regmaildb code could be pulled out into a separate "preparatory" commit.
In a way you will prepare existing code and transform it into a structure that can accept adding the new verifydb
stuff with minimal changes.

daemon/regmaild/regmaild.c
```diff
@@ -516,87 +704,204 @@ regcheck_ambiguous_id(const char *userid)
...
+    int8_t *rest = (int8_t *)&storage;
+    rest += sizeof(*header);
 
     assert(ev);
 
-    if ( (event & EV_TIMEOUT) ||
-        !(event & EV_READ) ||
-         toread(fd, &req, sizeof(req)) != sizeof(req) ||
-         req.cb != sizeof(req))
+    if ((event & EV_TIMEOUT) ||
```
wens: Personally I like the style of "matching the opening parenthesis", as it allows one to quickly see which parenthesis level it is part of.

daemon/regmaild/regmaild.c
```diff
@@ -516,87 +704,204 @@ regcheck_ambiguous_id(const char *userid)
...
+            toread(fd, header, sizeof(*header)) != sizeof(*header) ||
+            header->cb < sizeof(*header) ||
+            header->cb > sizeof(storage) ||
+            toread(fd, rest, header->cb - sizeof(*header)) != header->cb - sizeof(*header))
     {
-        fprintf(stderr, "error: corrupted request.\r\n");
-        close(fd);
-        free(ev);
-        return;
+        fprintf(stderr, "error: corrupted request.\n");
```
wens: Either leave the carriage return alone for now, or make a patch removing it from all log messages.
It is unrelated to what the commit log says the change is supposed to do.
(Why were there carriage returns in the first place...)

include/daemons.h
```diff
@@ -31,6 +56,60 @@ typedef struct
...
 }   regmaildb_req;
 
+#define VERIFYDB_VKEY_SIZE (160)
+
+// verifydb status codes.
+#define VERIFYDB_OK (0)
+#define VERIFYDB_ERROR (-1)
+
+// verifydb vmethod keys.
+#define VMETHOD_UNSET (0)
```
wens: Use an `enum`.

include/daemons.h
```diff
@@ -31,6 +56,60 @@ typedef struct
...
+
+typedef struct {
+    size_t cb;
```
wens, line 74: This deserves a better name. Based on your usage, the shortest unambiguous name I can think of is `bytecount`.

wens (commented):
First commit reviewed.
The commit message should be clear that it is introducing a framework, and does not actually support any vmethods at this point.

daemon/regmaild/regmaild.c
robertabcd: This will match if another user is currently using the requested `(vmethod, vkey)`.

daemon/regmaild/regmaild.c
robertabcd: I think this will become very hard to follow esp. in `verifydb_get`. As the performance isn't of concern here, `if` wouldn't hurt and I feel a single bail out point would be easier to follow.

daemon/regmaild/regmaild.c
robertabcd: I won't say which function but it will contain some details in `errmsg` which is enough to know which function call failed.

daemon/regmaild/regmaild.c
robertabcd: Done.

daemon/regmaild/regmaild.c
robertabcd: Separated out.

daemon/regmaild/regmaild.c
robertabcd: Done.

daemon/regmaild/regmaild.c
robertabcd: Made a separate commit that removes "\r".

include/daemons.h
robertabcd: Done.

include/daemons.h
robertabcd: This file uses `cb` across the board. I think we should keep the name for consistency purpose, unless we want to change them all.

daemon/regmaild/regmaild.c
```diff
@@ -526,34 +548,32 @@ client_cb(int fd, short event, void *arg)
...
     {
         case REGMAILDB_REQ_COUNT:
-            if (!*req.userid || !*req.email) {
-                err_request(&req, fd, ev);
-                return;
-            }
-            ret = regmaildb_check_email(req.email, strlen(req.email), req.userid);
+        {
+            if (!validate_regmaildb_request(&req))
+                goto end;
```
wens: I believe you can move this outside the switch / case.
After that you can either drop the default case, or mark it as not reached.

wens:
"regmaild: Error and exit if DB cannot be opened. " looks like it should come before "regmaild: Use g_Db in handler." to preserve working functionality during bisection.

include/daemons.h
```diff
@@ -31,6 +31,19 @@ typedef struct
...
+    int operation;
+} regmaildb_req_header;
+
+// Storage struct that fits all possible messages.
+typedef struct {
+    union {
+        regmaildb_req_header header;
+        regmaildb_req regmaildb;
+    } m;
+} regmaildb_req_storage;
```
wens, line 117: There might be a better way to do this:

```
typedef struct {
    size cb;
    int operation;
} regmaildb_req_header;

typedef struct {
    regmaildb_req_header; /* not sure if anonymous nested structs work this way */
    [request type-specific fields ... ]
} regmaildb_req;
```

But let's try that later.

daemon/regmaild/regmaild.c
robertabcd: Um, in later commit I added `validate_verifydb_request` and I would have to move them back. I'd like to just keep them like this.

include/daemons.h
robertabcd: This doesn't work. I notice I don't need to give a name to `union`, so I've removed it. This lets us do `req_storage.header` etc.

robertabcd:
> "regmaild: Error and exit if DB cannot be opened. " looks like it should come before "regmaild: Use g_Db in handler." to preserve working functionality during bisection.

Right, reordered.

robertabcd:
Accidentally merged two commits together. Reuploaded.

include/daemons.h
wens: Just curious, why doesn't it work though?

daemon/regmaild/regmaild.c
```diff
@@ -270,6 +276,202 @@ regmaildb_update_email(const char * userid, int userid_len, const char * email,
...
+
+    buf = evbuffer_new();
+    if (!buf)
+        goto end;
+    while (sqlite3_step(Stmt) == SQLITE_ROW) {
+        const int64_t timestamp = sqlite3_column_int64(Stmt, 0);
+        const int32_t vmethod = sqlite3_column_int64(Stmt, 1);
+        const char *vkey = (const char *)sqlite3_column_text(Stmt, 2);
+        if (!vkey)
+            continue;  // Bad record.
```
wens: Can't you just filter these out with the WHERE clause?
This way you could also get rid of `evbuffer_prepend()` and
instead queue up everything in order, or even drop `evbuffer` altogether.

daemon/regmaild/regmaild.c
```diff
@@ -270,6 +276,202 @@ regmaildb_update_email(const char * userid, int userid_len, const char * email,
...
+    if (r != SQLITE_OK)
+        goto end;
+
+    while (sqlite3_step(Stmt) == SQLITE_ROW) {
+        userec_t u = {};
+        const char *userid = (const char *)sqlite3_column_text(Stmt, 0);
+        int64_t generation = sqlite3_column_int64(Stmt, 1);
+
+        if (userid == NULL)
+            continue;  // Bad record.
```
wens, line 413: You can get rid of this case by checking for `userid IS NOT NULL` in the `WHERE` clause.

daemon/regmaild/regmaild.c
```diff
@@ -607,6 +849,47 @@ client_cb(int fd, short event, void *arg)
...
+        }
+
+        case VERIFYDB_REQ_GET:
+        {
+            const verifydb_req *req = &storage.verifydb;
+            if (!validate_verifydb_request(req))
+                goto end;
+
+            verifydb_get(req, fd);
+            goto end_noclose;
```
wens: Why the difference? Please keep all handling consistent.

wens:
For `regmailc: Add client support for verifydb.` please also split up the cleanup/restructure and verifydb additions.

wens:
Reviewed-by for the following:
- regmaild: Use "\n" as new line, instead of "\r\n".
- regmaild: Error and exit if DB cannot be opened.
- regmaild: Use g_Db in handler.
- regmaild: Factor out request validation into utility function.
- regmaild: Prepare for accepting different request struct per operation.
- regmaild: Return non-zero exit code on init error.
- regmaild: Refine error messages; nit style fixes.
- regmaild: Add initemaildb into list of programs.

Please push these out to master and rebase, so we don't have to look at them again.

For `Fix include in daemons.h.`, it looks like you squashed in something wrong.

mbbsd/register.c
```diff
@@ -1264,7 +1262,7 @@ create_regform_request()
 }
 
 static int
-register_email_input(const char *userid, char *email)
+register_email_input(const userec_t *u, char *email)
```
wens, line 1293: Within the commit `register: Move userid hack into emaildb functions.`, the parameter type change is unwarranted.
Please move this to the commit that actually needs it, which I assume is `register: Dual read/write to verifydb.`

wens:
Also, you don't need a full stop (period) at the end of the commit subject.
It is a subject line, much like for an article, not a full sentence, so the full stop is not needed.

daemon/regmaild/regmaild.c
robertabcd: My bad. It should return the record though. I made it return an empty string.
Right now, we won't have this case.

robertabcd: There may be multiple records. It needs the count in the reply struct that is going to be prepended.

daemon/regmaild/regmaild.c
robertabcd: Fixed.

mbbsd/register.c
robertabcd: Moved.

robertabcd:
> Also, you don't need a full stop (period) at the end of the commit subject.
> It is a subject line, much like for an article, not a full sentence, so the full stop is not needed.

Fixed.

include/daemons.h
robertabcd:
```c
typedef struct {
  regmaildb_req_header;
  int xxx;
} test_t;

void f() {
    test_t t;
    t.cb = 100;
}
```

```
../../include/daemons.h:120:3: warning: declaration does not declare anything [-Wmissing-declarations]
  regmaildb_req_header;
  ^~~~~~~~~~~~~~~~~~~~

regmaild.c:890:7: error: no member named 'cb' in 'test_t'
    t.cb = 100;
    ~ ^
```

wens: Ah, ok, so anonymous members don't work this way. Too bad.

robertabcd:
Discarding this. Working on a flatbuffers protocol.

PR #87, opened 2020-10-07
wsproxy: Substitute LuaJIT built-in library `ffi` for external library `vstruct`

IepIweidieng:
Recent OpenResty versions use LuaJIT since the `receiveany()` method was introduced.

LuaJIT has a built-in `ffi` library (<http://luajit.org/ext_ffi.html>), which can be used to replace the external `vstruct` library currently in use.

wsproxy was ported to DreamBBS (<https://term.ccns.cc>) (a Maple3-derived BBS), and these commits have been tested there (<https://github.com/ccns/dreambbs/commits/master/scripts/wsproxy>).

### DIFF
--- a/daemon/wsproxy/README.md
+++ b/daemon/wsproxy/README.md
@@ -2,10 +2,6 @@
 
 The websocket to telnet bbs proxy.
 
-## Dependency
-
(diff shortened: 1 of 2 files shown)


robertabcd:
Cool. I like it. Thank you for removing this dep.

PR #88, opened 2020-10-15
Fix `undeclared identifier` error when `USE_VERIFYDB` is not defined

IepIweidieng:
`mbbsd/menu.c` unconditionally references `verifydb_admin_search_display`, but the declaration of the function is disabled when `USE_VERIFYDB` is not defined.
This causes the compilation fails with the aforementioned error.

This fixes the problem by surrounding the referencing code with an appropriate preprocessor condition.


robertabcd:
Thanks!

PR #95, opened 2021-01-03
common/sys/string.c: Fix `strip_nonebig5` ignoring the byte following `ff fd`

### LINKED ISSUE #94: Possible one more i++ in cmsys.strip_nonebig5？
https://github.com/ptt/pttbbs/blob/master/common/sys/string.c#L283

0xFFFD happened in u2b.
We can paste thai characters in utf8-terminal
and get 0xFFFD.

strip_nonebig5 affects 
title in mbbsd/bbs.readdoent  and
currutmp->nickname in setup_utmp.

IepIweidieng:
This fixes #94.

According to `common/sys/big5.c`:
- Unicode code points without the corresponding Big5-UAO code point are always converted to `0xfffd` (see `u2b_table`).
- Big5-UAO code points `0xfeff`-`0xffff` are left unchanged when converted into Unicode (see `b2u_table`).

Consider the case for `strip_nonebig5`: `str` is `"\xff\xfdQ"` and `maxlen` > `3`.

When `i` is `0`, both `str[i]==255` and `251<=str[i+1] && str[i+1]<=254` hold, and the first `i++` following them skips the byte `\xff`.

`i+1<maxlen && str[i+1]` also holds, and the second `i++` skips the byte `\xfd`.

However, the `for` loop also performs `i++` after the `continue` and makes the byte `Q` ignored.

Therefore, the second `i++` is redundant and erroneous. It makes the byte following `"\xff\xfd"` ignored.

### DIFF
--- a/common/sys/string.c
+++ b/common/sys/string.c
@@ -276,15 +276,7 @@ strip_nonebig5(unsigned char *str, int maxlen)
   for(i=0;i<maxlen && str[i];i++) {
     if(32<=str[i] && str[i]<128)
       str[len++]=str[i];
-    else if(str[i]==255) {
-      if(i+1<maxlen)
-	if(251<=str[i+1] && str[i+1]<=254) {
-	  i++;
-	  if(i+1<maxlen && str[i+1])
-	    i++;
-	}
-      continue;
-    } else if(str[i]&0x80) {
+    else if(str[i]&0x80) {
       if(i+1<maxlen)
 	if((0x40<=str[i+1] && str[i+1]<=0x7e) ||
 	   (0xa1<=str[i+1] && str[i+1]<=0xfe)) {


kcwu:
查了才發現這段 code 一開始是我寫的 XD
https://github.com/ptt/pttbbs/commit/5707bbf7d35bd2e445114451e83e6519965079ce

當初那段邏輯(251<= ... <=254)的目的其實是要濾掉 telnet 的 IAC 
http://www.tsnien.idv.tw/Internet_WebBook/chap11/11-4%20Telnet%20%E9%80%9A%E8%A8%8A%E5%8D%94%E5%AE%9A.html

後來 pttbbs 已經可以正確處理 telnet protocol, 一般輸入欄位不會再拿到 IAC sequence 了, 所以我想 `if(str[i]==255)` 那段邏輯應該可以整個砍掉了.

wens:
原本的做法看起來也不是很正確XDD

IepIweidieng:
As @kcwu pointed out, `strip_nonebig5` was meant to strip telnet commands (RFC 854).

Therefore `"\xff\xfdQ"` will be misinterpreted as telnet command `IAC DO 0x51` and stripped, which makes the byte `Q` ignored.

For addressing this issue, I have force pushed to update the commit to drop the code which strips the telnet commands.

kcwu:
LGTM

wens:
Could you expand the first part of your commit log a bit more?
The following is what I had written. You can see that I include more context
about why `strip_nonebig5()` strips out telnet IAC characters. Please also
include the commit references.

-------------------------------------------------------------------

strip_nonebig5() was originally strip_iac(), which filtered out telnet
protocol IAC bytes to avoid confusing the client. See commit 7b2c3745
("strip IAC char in user's nick")

It was later expanded to filter out non-Big5 codepoints to avoid messing
up displays. The original IAC-filtering code was kept. See commit
5707bbf7 ("strip none-big5 char in user's nick").

Since we now have a dedicated filter layer that handles the telnet
protocol, which includes inserting extra 0xff bytes to avoid 0xff
in Big5 characters being interpreted as IAC, we can drop this bit
of code, which has caused some legal Big5 characters being filtered
out.

IepIweidieng:
> Could you expand the first part of your commit log a bit more?
> The following is what I had written. You can see that I include more context
> about why `strip_nonebig5()` strips out telnet IAC characters. Please also
> include the commit references.
> 
> strip_nonebig5() was originally strip_iac(), which filtered out telnet
> protocol IAC bytes to avoid confusing the client. See commit [7b2c374](https://github.com/ptt/pttbbs/commit/7b2c37455c466e5304a0251502533cdd78ab19ab)
> ("strip IAC char in user's nick")
> 
> It was later expanded to filter out non-Big5 codepoints to avoid messing
> up displays. The original IAC-filtering code was kept. See commit
> [5707bbf](https://github.com/ptt/pttbbs/commit/5707bbf7d35bd2e445114451e83e6519965079ce) ("strip none-big5 char in user's nick").
> 
> Since we now have a dedicated filter layer that handles the telnet
> protocol, which includes inserting extra 0xff bytes to avoid 0xff
> in Big5 characters being interpreted as IAC, we can drop this bit
> of code, which has caused some legal Big5 characters being filtered
> out.

Sure!
I have force pushed to update the commit to include the information.

wens:
I manually picked the commit and cleaned up some issues with the commit log,
including indentation, ordering of paragraphs, and how some identifiers are
presented.

IepIweidieng:
After further investigation, I found out that `IAC WILL/WONT/DO/DONT <option_code>` is filtered out by `strip_nonebig5` only after 5707bbf7d35bd2e445114451e83e6519965079ce.

I have force pushed to update the commit to reflect this finding.

wens:
Please try not to use too much GitHub formatting in commit logs, as they are useless when viewed from a console.
Also, it would be nice if you could use your real full name and email in the future. This is standard practice in many open source projects.

IepIweidieng:
> Please try not to use too much GitHub formatting in commit logs, as they are useless when viewed from a console.
> Also, it would be nice if you could use your real full name and email in the future. This is standard practice in many open source projects.

Thanks for the suggestion!

wens:
> After further investigation, I found out that `IAC WILL/WONT/DO/DONT <option_code>` is filtered out by `strip_nonebig5` only after [5707bbf](https://github.com/ptt/pttbbs/commit/5707bbf7d35bd2e445114451e83e6519965079ce).
> 
> I have force pushed to update the commit to reflect this finding.

I updated the commit log and forced pushed the master branch.

IepIweidieng:
> After further investigation, I found out that `IAC WILL/WONT/DO/DONT <option_code>` is filtered out by `strip_nonebig5` only after [5707bbf](https://github.com/ptt/pttbbs/commit/5707bbf7d35bd2e445114451e83e6519965079ce).
> 
> I have force pushed to update the commit to reflect this finding.

FYI, 7b2c37455c466e5304a0251502533cdd78ab19ab made `strip_nonebig5` simply strip all `0xff` data bytes.

`common/sys/telnet.c` seems to be developed later than 7b2c37455c466e5304a0251502533cdd78ab19ab, but I am unsure whether the filtering of telnet commands in 7b2c37455c466e5304a0251502533cdd78ab19ab became an issue only after `common/sys/telnet.c` was developed.

Issue #100, opened 2021-02-03
whence in PttLock should be SEEK_SET? (or PttLock after lseek should set offset as 0?)


chhsiao1981:
https://github.com/ptt/pttbbs/blob/master/common/sys/lock.c#L15

Found several lines using PttLock while already did lseek.
(but some didnt use lseek before PttLock)

Not exactly sure whether offset of lseek is the same as the fnctl,
and not exactly sure whether fcntl F_SETLKW affects seek-pointer though.

The following lines use PttLock F_WRLCK

https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L84 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L133 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L195 (no lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L261 (no lseek)

The following lines use PttLock F_UNLCK
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L86 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L149 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L223 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L275 (after lseek)

chhsiao1981:
strangely there are lseek between F_WRLCK / F_UNLCK in the following functions:

record.substitute_record2
record.delete_records
record. delete_record2

Not exactly sure whether it's just that I didn't fully know fnctl / lseek mechanism,
or we do need to change to SEEK_SET in PttLock.

chhsiao1981:
The reason why record.substitute_record2 works is because lseek is SEEK_SET to same offset before F_WRLCK

The reason why record.delete_records, record.delete_record2 work is because the fd is closed after unlock.
(lock is released and no deadlock.)

wens:
Based on the usage patterns you listed, which covers about everything, we should probably be using SEEK_SET in PttLock.

chhsiao1981:
resolved in #101

PR #101, opened 2021-02-03
Provided offset, we should use SEEK_SET in PttLock

chhsiao1981:
Found some functions use PttLock after lseek
but some use PttLock without lseek.
Provided offset, we should use SEEK_SET in PttLock.

The following lines use PttLock F_WRLCK:
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L84 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L133 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L195 (no lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L261 (no lseek)

The following lines use PttLock F_UNLCK:
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L86 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L149 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L223 (after lseek)
https://github.com/ptt/pttbbs/blob/master/common/sys/record.c#L275 (after lseek)

https://github.com/ptt/pttbbs/issues/100

### DIFF
--- a/common/sys/lock.c
+++ b/common/sys/lock.c
@@ -7,13 +7,13 @@
  * @param mode	F_WRLCK, F_UNLCK
  */
 void
-PttLock(int fd, int start, int size, int mode)
+PttLock(int fd, int offset, int size, int mode)
 {
     static struct flock lock_it;
     int             ret;
 
-    lock_it.l_whence = SEEK_CUR;/* from current point */
-    lock_it.l_start = start;	/* -"- */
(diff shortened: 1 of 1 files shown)


hungte:
It is very surprising how we had the code running so long without doing lock correctly

PR #107, opened 2022-03-08
Supporting Ban Post or Comment outside Taiwanese IP

PichuChen:
Requirement:
```
- 新增看板權限限制非台灣IP發/推文 (實作難度: 易)
  Please use the country value returned from `fromd`.
  Please guard this via a compile time flag.
```

Changes:
1. Introduce a new compiler flag `POST_COMMENT_WITH_TAIWAN_IP_ONLY` for this function.
2. Implement `is_taiwan_ip()` which determine if current user is from Taiwan IP.
3. Insert checking Taiwan IP into `const char * postperm_msg(const char *bname)`.
4. Introduce a board configuration `BRD_TAIWAN_IP_ONLY (0x2000000)` for this function.

Screenshots:
截圖 2022-03-08 下午8 04 53
截圖 2022-03-08 下午8 05 12
截圖 2022-03-08 下午8 05 50

Remark:
* Please enable `FROMD` and flag `POST_COMMENT_WITH_TAIWAN_IP_ONLY` when use this function.

TODO:
* Changing `from_cc` to ISO code instead of Big-5 string.
* Review the post or comment checking flow and write the document.

### DIFF
--- a/include/proto.h
+++ b/include/proto.h
@@ -101,6 +101,9 @@ void log_crosspost_in_allpost(const char *brd, const fileheader_t *postfile);
 #ifdef USE_COOLDOWN
 int check_cooldown(boardheader_t *bp);
 #endif
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+bool is_taiwan_ip();
+#endif
 
 /* board */
 #define setutmpbid(bid) currutmp->brc_id=bid;
--- a/mbbsd/bbs.c
+++ b/mbbsd/bbs.c
@@ -433,6 +433,23 @@ int IsFreeBoardName(const char *brdname)
     return 0;
 }
 
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+/*
+ * Check if user ip address is from Taiwan
+ *
+ * This function will compare from_cc string and "Taiwan" in Chinese (Big5)
+ * from_cc will generate when user login bbs system, and set from FROMD
+ *
+ * Return true when user is from Taiwan IP, false when outside Taiwan
+ */
+bool
+is_taiwan_ip()
+{
+    // TODO: move this string to string table, eg: common.h
+    return !strncmp("»OÆW", from_cc, sizeof(from_cc));
+}
+#endif
+
 int
 CheckModifyPerm(const char **preason)
 {
--- a/mbbsd/board.c
+++ b/mbbsd/board.c
@@ -496,6 +496,13 @@ b_config(void)
 		(bp->brdattr & BRD_OVER18) ?
 		ANSI_COLOR(1) "禁止 " : "允許\ " );
 
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+	prints( " " ANSI_COLOR(1;36) "t" ANSI_RESET
+		" - %s" ANSI_RESET "台灣IP發/推文\n",
+		(bp->brdattr & BRD_TAIWAN_IP_ONLY) ?
+		ANSI_COLOR(1) "限定 " : "不限定 " );
+#endif
+
 	if (!canpost)
 	    outs(ANSI_COLOR(1;31)"  ★ 您在此看板無發文或推文權限，"
 		"詳細原因請參考上面顯示為紅色或有 * 的項目。"ANSI_RESET"\n");
@@ -812,6 +819,13 @@ b_config(void)
 		    touched = 1;
 		}
 		break;
+
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+		case 't':
+		bp->brdattr ^= BRD_TAIWAN_IP_ONLY;
+		touched = 1;
+		break;
+#endif // POST_COMMENT_WITH_TAIWAN_IP_ONLY
 
 	    case 'v':
 		clear();
--- a/daemon/fromd/Makefile
+++ b/daemon/fromd/Makefile
@@ -15,7 +15,8 @@ FROMD_LDFLAGS = -levent
 .if ${ENABLE_MAXMIND_DB} == "y"
 CFLAGS += -DFROMD_USE_MAXMIND_DB
 FROMD_LDFLAGS += -lmaxminddb
-.elif ${ENABLE_GEOIP_DB} == "y"
+.endif
+.if ${ENABLE_GEOIP_DB} == "y"
 CFLAGS += -DFROMD_USE_GEOIP_DB
 FROMD_LDFLAGS += -lGeoIP
 .endif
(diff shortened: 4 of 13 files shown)


mbbsd/board.c
```diff
@@ -496,6 +496,15 @@ b_config(void)
 		(bp->brdattr & BRD_OVER18) ?
 		ANSI_COLOR(1) "�T�� " : "���\\ " );
 
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+	// use '8' instead of '1', to prevent 'l'/'1' confusion
```
kcwu: incorrect comment

kcwu:
please squash your 4 commits into single one

mbbsd/board.c
PichuChen: fixed, thanks

include/proto.h
```diff
@@ -101,6 +101,9 @@ void log_crosspost_in_allpost(const char *brd, const fileheader_t *postfile);
 #ifdef USE_COOLDOWN
 int check_cooldown(boardheader_t *bp);
 #endif
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
+int is_taiwan_ip();
```
robertabcd: Please return `bool`.

mbbsd/bbs.c
```diff
@@ -433,6 +433,26 @@ int IsFreeBoardName(const char *brdname)
...
+ * Check if user ip address is from Taiwan
+ *
+ * This function will compare from_cc string and "Taiwan" in Chinese (Big5)
+ * from_cc will generate when user login bbs system, and set from FORMD
+ *
+ * Return 1 when user is from Taiwan IP, 0 when outside Taiwan
+ */
+int is_taiwan_ip()
+{
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
```
robertabcd: Might just wrap this `#ifdef` around the whole function, so you don't need the `#else` part.

PichuChen: Let me think twice. If I warp whole function, I will need to change the function name or the future developer might use this function without `POST_COMMENT_WITH_TAIWAN_IP_ONLY` flag and got compiler error unexpected.

robertabcd: That should be a compiler error, right?

I don't think it's correct that `is_taiwan_ip()` always return `true` without `POST_COMMENT_WITH_TAIWAN_IP_ONLY`.

You also have a ifdef in the header file already.

PichuChen: Yes, you're right.

mbbsd/bbs.c
```diff
@@ -433,6 +433,24 @@ int IsFreeBoardName(const char *brdname)
     return 0;
 }
 
+/*
+ * Check if user ip address is from Taiwan
+ *
+ * This function will compare from_cc string and "Taiwan" in Chinese (Big5)
+ * from_cc will generate when user login bbs system, and set from FORMD
```
kcwu: s/FORMD/FROMD/ ?

kcwu: BTW, you have the same typo in the description of this pull request as well.

PichuChen: Thank you 😣

mbbsd/bbs.c
```diff
@@ -433,6 +433,24 @@ int IsFreeBoardName(const char *brdname)
...
+/*
+ * Check if user ip address is from Taiwan
+ *
+ * This function will compare from_cc string and "Taiwan" in Chinese (Big5)
+ * from_cc will generate when user login bbs system, and set from FROMD
+ *
+ * Return true when user is from Taiwan IP, false when outside Taiwan
+ */
+
+#ifdef POST_COMMENT_WITH_TAIWAN_IP_ONLY
```
kcwu: IMO, the function comment should be inside the #ifdef block.

mbbsd/bbs.c
```diff
@@ -433,6 +433,24 @@ int IsFreeBoardName(const char *brdname)
...
+bool
+is_taiwan_ip()
+{
+    // TODO: move this string to string table, eg: common.h
+    return !strncmp("�O�W", from_cc, STRLEN);
```
kcwu: Given from_cc is properly null terminated, use strcmp instead of strncmp.
If you don't trust the null character and want to be extremely safe, use sizeof(from_cc) instead of STRLEN, just in case the size may be changed in the future.

PichuChen: Agree, using `sizeof(from_cc)` is better.

mbbsd/bbs.c
PichuChen: Sure.

IepIweidieng:
*Disclaimer: I'm not directly associated with any of the main maintainers of this repository.*

I'm not sure whether directly put `taiwan` in the name of functions, macros, etc. is elegant for implementing region restriction.
Displaying "Taiwan"/"臺灣" as the main region to the users of Ptt seems totally fine, however.

I think that it might be better if a work is done to add an option in `config.h` or `pttbbs.conf` which controls the region exempted from the restriction.
Assume the option will be named `STR_FROM_REGION_LOCAL` (or `ISO_CODE_REGION_LOCAL` and the likes). As an example, a server maintainer willing to change the exempted region from Taiwan to America can write `#define STR_FROM_REGION_LOCAL "美國"` (default: `"臺灣"`) in `pttbbs.conf` .

Although this is not an issue until some person or organization is willing to establish a BBS targeting people oversea using the PttBBS server code, so perhaps such work deserves another PR.

PichuChen:
> _Disclaimer: I'm not directly associated with any of the main maintainers of this repository._
> 
> I'm not sure whether directly put `taiwan` in the name of functions, macros, etc. is elegant for implementing region restriction. Displaying "Taiwan"/"臺灣" as the main region to the users of Ptt seems totally fine, however.
> 
> I think that it might be better if a work is done to add an option in `config.h` or `pttbbs.conf` which controls the region exempted from the restriction. Assume the option will be named `STR_FROM_REGION_LOCAL` (or `ISO_CODE_REGION_LOCAL` and the likes). As an example, a server maintainer willing to change the exempted region from Taiwan to America can write `#define STR_FROM_REGION_LOCAL "美國"` (default: `"臺灣"`) in `pttbbs.conf` .
> 
> Although this is not an issue until some person or organization is willing to establish a BBS targeting people oversea using the PttBBS server code, so perhaps such work deserves another PR.

I am just follow the requirement, but in technically, I will design for an IP address white list or black list, or an ASN white list or black list.

Btw, the requirement has been changed, they want to check (or ban) the user according there AOTP register status, so you can see this PR (https://github.com/ptt/pttbbs/pull/108)

hungte:
First of all, it is better to implement this as region-specific (like @IepIweidieng mentioned) feature instead of hard-coding "Taiwan".

Second, given the author said "the requirement has been changed", I'm going to close the PR.

PR #109, opened 2022-04-17
Improve contact email

chhsiao1981:
Currently we can change "contact email" without any "re-verification".
This leads the bad-guys still able to arbitrarily change the contact email once they are able to login,
obsoleting the original purpose of the "contact email".

The original reason that we don't do re-verification 
was because the users already lost "verify-email",
unable to verify the 1st-time contact email,
especially that "verify-email" is restrictive.

I would argue that we don't need to do "re-verify" for the 1st-time "contact email",
but we do want to do "re-verify" once the email becomes "contact email".

This PR add the followings to reflect the change:
1. add is_contact_email in userec_t.
2. add challenge-email to original contact-email if changing the contact-email again.
3. add migrate_is_contact_email.
4. add check_contact_email (to remind the users update the contact-email.)


chhsiao1981:
(close because it doesn't fit current PTT policy.)

PR #110, opened 2022-04-21
add check_contact_email (contact_whitemail, contact_banemail)

chhsiao1981:
add check_contact_email (contact_whitemail, contact_banemail)

### DIFF
--- a/mbbsd/register.c
+++ b/mbbsd/register.c
@@ -5,6 +5,7 @@
 #define FN_REGISTER_LOG  "register.log"	// global registration history
 #define FN_REJECT_NOTIFY "justify.reject"
 #define FN_NOTIN_WHITELIST_NOTICE "etc/whitemail.notice"
+#define FN_NOTIN_CONTACT_WHITELIST_NOTICE "etc/contact_whitemail.notice"
 
 // Regform1 file name (deprecated)
 #define fn_register	"register.new"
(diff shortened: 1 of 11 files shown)


chhsiao1981:
* separate check_email_allow_reject_lists to be check_email_allow_reject_lists and check_email_allow_reject_lists_core
   (check_email_allow_reject_lists_core will be reused with check_contact_email_allow_reject_lists)
* implement check_contact_email by copying check_regmail, but without ein->allow_untrusted
   (It's possible that we will have different methods for check_regmail and check_contact_email in the future. I think it's better to write them separately for now.)
* Although the UX is not quite good in that the user needs to do the reg-code first and then gets rejected, but only bad guys get rejected. We don't prioritize the UX for now.

wens:
Please write proper commit messages. GitHub is not a substitute for this.

mbbsd/register.c
```diff
@@ -70,6 +70,9 @@ register_count_email(const userec_t *u, const char *email);
 static int
 register_check_and_update_emaildb(const userec_t *u, const char *email);
 
+bool
```
wens: static

mbbsd/register.c
```diff
@@ -70,6 +70,9 @@ register_count_email(const userec_t *u, const char *email);
...
+check_email_allow_reject_lists_core(char *email, const char **errmsg, const char **notice_file, const char *white_email, const char *ban_email);
```
wens: You are not using this function before it's definition. Why add a forward declaration?

chhsiao1981: It's just my personal preference (to easily know what are the functions in the file by reading the top of the files / header files.) I'll remove it.

chhsiao1981: actually check_email_allow_reject_lists_core is defined at line: 1021, but is used by check_email_allow_reject_lists at line: 1017.

It's my personal preference to minimize the diff with the original check_email_allow_reject_lists.

I can move check_email_allow_reject_lists after check_email_allow_reject_lists_core to remove the declaration.

wens:
Please rebase and clean up your changes. Each commit should be one logical change.

chhsiao1981:

PR #111, opened 2022-04-22
mbbsd: notify user after resetting password in user configuration

chhsiao1981:
1. move AccountRecovery::NotifyUser to passwdnotify::PasswdNotify::NotifyUser.
2. notify user after resetting password in user configuration (in addition to AccountRecovery).
3. add ip in the subject add add how-to-reset-passwd link in the template.

### DIFF
--- a/include/passwd_notify.hpp
+++ b/include/passwd_notify.hpp
@@ -1,0 +1,9 @@
+#pragma once
+
+#include <string>
+
+namespace passwdnotify {
+
+void NotifyUser(const std::string &userid, const std::string &fromhost, const std::string &email);
+
+} //namespace
--- a/mbbsd/passwd_notify.cc
+++ b/mbbsd/passwd_notify.cc
@@ -1,0 +1,28 @@
+extern "C" {
+#include "bbs.h"
+#include "daemons.h"
+}
+
+#include "passwd_notify.hpp"
+
+#include <string>
+
+namespace passwdnotify {
+
+void NotifyUser(const std::string &userid,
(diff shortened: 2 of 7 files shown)


include/passwd_notify.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <string>
+
+namespace passwdnotify {
+
+class PasswdNotify {
```
wens: Why would you even put in a class if there's no state...

chhsiao1981: It's my personal preference in C++/Java (to easily know which functions are public functions).

I'll remove the class statement.

wens: If it's not static, it's globally visible...

wens:
Please rebase and cleanup your changes...

mbbsd/passwd_notify.cc
```diff
@@ -0,0 +1,28 @@
...
+
+void NotifyUser(const std::string &userid,
+                              const std::string &fromhost,
+                              const std::string &email) {
+  std::string subject;
+  subject.append(" " BBSNAME " - ");
+  subject.append(userid);
+  subject.append("(");
+  subject.append(fromhost);
+  subject.append("), �z���K�X�w����");
```
chhsiao1981, line 20: This is due to github encoding issue. The original file should be ok.

wens:
Ended up reimplementing it in C to avoid pulling in C++ dependencies just for one function.

PR #113, opened 2022-04-26
mbbsd: remind setting contact-email after successfully login

chhsiao1981:
mbbsd: remind setting contact-email after successfully login (in user_login)

mbbsd: add check_contact_email
util: add migrate_is_contact_email

Currently the contact-email is mixed with the register-email, letting us unable
to know whether the email is contact-email or just the register-email.

Adding 1-byte-flag (uint8_t) is_contact_email in userec_t to know whether the email
is contact-email or not.

Setup the flag as long as we change the contact-email (even if the email is the same).

Previously we don't set is_contact_email in userec_t, but we have the logs in user.security.
Use user.security to migrate is_contact_email.

To deploy this update:
1. make all (no install)
2. do migrate_is_contact_email from /home/bbs
3. make install

To discuss:
whether we should increase the SHM version?
considering that there is no real break change with the original version.

### DIFF
--- a/daemon/fromd/Makefile
+++ b/daemon/fromd/Makefile
@@ -15,7 +15,8 @@ FROMD_LDFLAGS = -levent
 .if ${ENABLE_MAXMIND_DB} == "y"
 CFLAGS += -DFROMD_USE_MAXMIND_DB
 FROMD_LDFLAGS += -lmaxminddb
-.elif ${ENABLE_GEOIP_DB} == "y"
+.endif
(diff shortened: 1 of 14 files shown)


wens:
Design change: if `email` field in `userec_t`, ask user to go through contact email setup procedure upon login. The user shall not be allowed to proceed if email is not set up.

chhsiao1981:
@wens 

Thanks for the comment. I'll close this PR due to this PR does not fit Design change anymore.

PR #114, opened 2022-05-04
mbbsd: add email_challenge

chhsiao1981:
mbbsd: add email_challenge

Currently we verify email again only in AccountRecovery.
We should verify email as well when the user intends to reset the password.

It is possible that we will need to verify the email when changing
the contact-email in the future.
Use email_challenge instead of passwd_challenge for this generalization.

The implementation is based on AccountRecovery::EmailCodeChallenge.

It is better to review #113 before merging this PR,
in that #113 enables us differentiating register-email vs. contact-email.

### DIFF
--- a/include/email_challenge.hpp
+++ b/include/email_challenge.hpp
@@ -13,11 +13,11 @@ extern "C" {
 
 #include "user_handle.hpp"
 
-void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails_);
-bool LoadVerifyDbEmail(const std::optional<UserHandle> &user_,
-                       std::vector<std::string> &all_emails_);
-void EmailCodeChallenge(const std::string &email_,
-                        const std::vector<std::string> &all_emails_,
-                        int &y_);
+void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails);
+bool LoadVerifyDbEmail(const std::optional<UserHandle> &user,
+                       std::vector<std::string> &all_emails);
+void EmailCodeChallenge(const std::string &email,
+                        const std::vector<std::string> &all_emails,
+                        int &y);
 
 #endif
--- a/include/email_challenge.hpp
+++ b/include/email_challenge.hpp
@@ -1,0 +1,23 @@
+#pragma once
+
+extern "C" {
+#include "bbs.h"
+#include "daemons.h"
+}
+
+#ifdef USE_VERIFYDB_ACCOUNT_RECOVERY
+
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "user_handle.hpp"
+
+void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails_);
+bool LoadVerifyDbEmail(const std::optional<UserHandle> &user_,
+                       std::vector<std::string> &all_emails_);
+void EmailCodeChallenge(const std::string &email_,
+                        const std::vector<std::string> &all_emails_,
+                        int &y_);
+
+#endif
--- a/include/email_challenge.hpp
+++ b/include/email_challenge.hpp
@@ -5,19 +5,23 @@ extern "C" {
 #include "daemons.h"
 }
 
-#ifdef USE_VERIFYDB_ACCOUNT_RECOVERY
-
 #include <optional>
 #include <string>
 #include <vector>
 
 #include "user_handle.hpp"
 
-void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails);
-bool LoadVerifyDbEmail(const std::optional<UserHandle> &user,
+namespace email_challenge {
+
+void LoadUserEmail(const userec_t *u, std::vector<std::string> &all_emails);
+bool LoadVerifyDbEmail(const std::optional<user_handle::UserHandle> &user,
                        std::vector<std::string> &all_emails);
-void EmailCodeChallenge(const std::string &email,
+void EmailCodeChallenge(const bool check_email,
+                        const std::string &email,
                         const std::vector<std::string> &all_emails,
+                        const std::string &prompt,
+                        const std::string &ip,
+                        const std::string &filename,
                         int &y);
 
-#endif
+} // namespace email_challenge
--- a/include/proto.h
+++ b/include/proto.h
@@ -228,6 +228,15 @@ void auto_backup(void);
 void restore_backup(void);
 const char *ask_tmpbuf(int y);
 
+/* email-challenge */
+int email_code_challenge(const char *email,
+                         const userec_t *u,
+                         const int y,
+                         const char *prompt,
+                         const char *ip,
+                         const char *filename,
+                         int *out_y);
+
 /* emaildb */
 #ifdef USE_EMAILDB
 int emaildb_check_email (const char *userid, const char *email);
@@ -580,7 +589,7 @@ void check_register(void);
 bool check_email_allow_reject_lists(
     char *email, const char **errmsg, const char **notice_file);
 void register_mail_complete_and_exit();
-int change_contact_email(bool skip_same_email_check);
+int change_contact_email(bool skip_same_email_check, bool require_two_factor_auth);
 
 /* register_sms */
 void u_sms_verification();
--- a/mbbsd/email_challenge.cc
+++ b/mbbsd/email_challenge.cc
@@ -32,19 +32,32 @@ NormalizeEmail(const std::string &email) {
   return out;
 }
 
-void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails_) {
+// LoadUserEmail
+//
+// Params:
+//   u:          userec
+//   all_emails: all valid emails.
+void LoadUserEmail(const userec_t &u, std::vector<std::string> &all_emails) {
 #ifdef USEREC_EMAIL_IS_CONTACT
   auto email = NormalizeEmail(u.email);
   if (email)
-    all_emails_.push_back(email.value());
+    all_emails.push_back(email.value());
 #endif
 }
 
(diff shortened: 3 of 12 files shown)


mbbsd/email_challenge.cc
```diff
@@ -0,0 +1,203 @@
...
+      if (each_email == normalized_email) {
+        email_matches = true;
+      }
+    }
+  }
+
+  if (email_matches) {
+    // We only want to send email if the user input the matching address.
+    // Silently fail if email is not matching, so that user cannot guess other
+    // user's email address.
```
wens: I'm not sure what you want to achieve with this. The user's email is already visible through the user info menu, even the email stored in the verification database.

chhsiao1981: I wrote email_challenge.cc based on recovery.cc

This is just trying to minimize the difference between recovery.cc and email_challenge.cc:

https://github.com/ptt/pttbbs/blob/master/mbbsd/recover.cc#L227

chhsiao1981: `all_emails` includes the contact email and the "original verified email".

It is possible that some user's contact email got updated by some bad guy, 
but the user can still use the "original verified email" to do the account-related operations (retrieve the account / change the contact-email again.)

chhsiao1981: `all_emails` includes "contact email" and "original verified email".

It's possible that some user's contact email got updated by some bad guy,
but the user can still use "original verified email" to do account-related operations (retrieve the account / change the email.)

wens: What I meant was the original verified email is visible via (U)ser -> (I)nfo -> (V), so the "prevent bad user from guessing" is moot.

Plus you already copied the implementation. What's the point of minimizing the difference? It's not like you are reusing it.

chhsiao1981: I'm sorry that I was not clear earlier～

1. The purpose of "email-challenge" is to send the challenge to the email, not just "guessing the verified email".
    From my understanding, now we have "contact-email" and "verified-email".
    There is only 1 "contact-email" and currently it can be easily modified.
    "verified-email" is harder to be modified because it requires valid verifiable email-address.

    It is possible that some user's contact-email got changed by the bad guy,
    but the user still can access the verified email.
    the user would like to recover the password / contact email through the verified email in this case.

2. I did want to merge the two functions. After all it should be the single function doing email-challenge.
    However, I'm not sure whether it is ok to change `AccountRecovery::EmailCodeChallenge` as well in this PR.
    So I just leave `AccountRecovery::EmailCodeChallenge` intact in this PR.

    I am very happy to merge the two functions in this PR if it is ok to do so.

wens: Re: 1.  But this is a separate function, i.e. 2FA _before_ changing password / email, so the user (legit or otherwise) already has access to said info. Why bother having the user input it again? Just send the challenge email to both the verified email and contact email.

Re: 2. Either merge them (which I doubt would work or look pretty), or just write functions that suite your need. Duplicating it is the worst choice.

chhsiao1981: @wens 

Thanks for the reply.

1. I didn't think of this alternative setup. I'll update to this setup.
2. I'll merge these two functions to simplify the code.

chhsiao1981:
This PR requires email-challenge only during "resetting password".

Therefore, we don't need to require resetting contact-email before approving this PR.

The potential requiring email-challenge when changing contact-email will be in other PR
(after approving this PR.)

chhsiao1981:
hi @wens ～

I believe that I followed your guideline and finished updating this PR.

USE_2FA with resetting-password / changing contact-email is implemented
in this PR as well.

Can you take a look again？～

Thanks!

mbbsd/register.c
```diff
@@ -1743,15 +1745,29 @@ static int notify_email_change(const char *userid, const char *email)
...
+//     1. do 2FA.
 //
 // Return:
 //   int: 0: ok -1: err
 int
-change_contact_email(bool skip_same_email_check)
+change_contact_email(bool skip_same_email_check, bool two_factor_authentication)
 {
     char email[EMAILSZ] = {};
     memcpy(email, cuser.email, sizeof(email));
```
wens, line 1767: I think we should force a reload from disk, i.e. call `pwcuReload()`, before this, in case another login instance changed the email. This should be a new precursor commit.

include/email_challenge.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "user_handle.hpp"
+
+namespace emailchallenge {
+
+bool EmailChallenge(bool check_input_email, const std::string &input_email, const userhandle::UserHandle &user, const int y, const std::string &prompt, const std::string &ip, const std::string &filename, int *out_y);
```
wens: Could you wrap the line at 80 or 100 characters, and align the following lines with the open parentheses?

include/proto.h
```diff
@@ -228,6 +228,9 @@ void auto_backup(void);
 void restore_backup(void);
 const char *ask_tmpbuf(int y);
 
+/* email-challenge */
+int email_challenge(const char *email, const userec_t *user, const int y, const char *prompt, const char *ip, const char *filename, int *out_y);
```
wens: Same here.

wens (commented):
Can you split this into two commits? The first commit only moves code and adds the namespace. The second one modifies the code to add whatever you added.

This would make it much more easier to diff and review.

chhsiao1981:
@wens 

Thanks for the comments!
I'll update it asap～

chhsiao1981:
hi @wens ～

Thank you so much for the comments~
I believe that I followed your guideline and finished updating this PR.

Can you take a look again？～

Thanks!

wens:
Please move [mbbsd: resync cuser.email in the beginning of change_contact_email](https://github.com/ptt/pttbbs/pull/114/commits/a9c153b5d30203338203404d9df5ac7a222d1935) to be the first, i.e. oldest, commit in the PR. Also please drop the period at the end.

mbbsd/email_challenge.cc
```diff
@@ -0,0 +1,270 @@
...
+#include <optional>
+#include <string>
+
+#include "verifydb.h"
+#include "verifydb.fbs.h"
+
+#include "user_handle.hpp"
+#include "email_challenge.hpp"
+
+namespace emailchallenge {
```
wens: Please add underscore between the two words. That makes it easier to read.

include/proto.h
```diff
@@ -580,7 +589,7 @@ void check_register(void);
 bool check_email_allow_reject_lists(
     char *email, const char **errmsg, const char **notice_file);
 void register_mail_complete_and_exit();
-int change_contact_email(bool skip_same_email_check);
+int change_contact_email(bool skip_same_email_check, bool two_factor_authentication);
```
wens: Better name for second parameter: `require_two_factor_auth`

mbbsd/email_challenge.cc
```diff
@@ -3,6 +3,12 @@ extern "C" {
 #include "daemons.h"
 }
 
+#ifdef USE_2FA
```
wens: Call this `USE_EMAIL_2FA` or something, in case later we have other forms of 2FA.

mbbsd/email_challenge.cc
```diff
@@ -3,6 +3,12 @@ extern "C" {
...
+# ifndef USE_VERIFYDB_ACCOUNT_RECOVERY
+#   error "USE_2FA requires USE_VERIFYDB_ACCOUNT_RECOVERY and USE_VERIFYDB"
```
wens: VERIFYDB is not strictly needed. You can always skip `LoadVerifyDbEmail()` and rely solely on user's listed email.

mbbsd/user_handle.cc
```diff
@@ -0,0 +1,34 @@
+extern "C" {
+#include "bbs.h"
+}
+
+#include <string>
+#include "user_handle.hpp"
+
+namespace userhandle {
```
wens: Add an underscore between the words.

mbbsd/email_challenge.cc
```diff
@@ -0,0 +1,270 @@
...
+static bool LoadVerifyDbEmail(const userhandle::UserHandle &user, std::vector<std::string> &all_emails);
+static bool AppendNonRepeatedEmail(const std::string &email, std::vector<std::string> &all_emails);
+static std::optional<std::string> NormalizeEmail(const std::string &email);
+static bool SendChallengeCode(const std::string &email,
+                              const std::string &code,
+                              const std::string &prompt,
+                              const std::string &ip,
+                              const std::string &filename);
+static std::string GenCode(size_t len);
+static void UserErrorExit();
```
wens: No need to forward declare static stuff. Just order them properly.

wens:
> Can you split this into two commits? The first commit only moves code and adds the namespace.
You should keep changes to a minimal. That includes any changes to comments and stuff. It should be a pure code move, with only necessary changes to make it compile.

If you use `git show <commit>--color-moved --color-moved-ws=ignore-all-space`, it should mark most of the commit as moved, and only bits where you added namespace as deletions and additions. You could even forgo the namespace and just have it in the global namespace.

wens:
Simply put, if you need to make modifications so the code fits, do them in separate commits. A code move is just a code move, there should not be any other changes. One commit equals one logical commit, without breaking the build.

chhsiao1981:
hi @wens ～

Thank you so much for your comments!
I'll update accordingly asap～

chhsiao1981:
hi @wens ～

Thank you so much for the comments~
I believe that I followed your guideline and finished updating this PR.

Can you take a look again？～

Thanks!

chhsiao1981:

PR #115, opened 2022-05-26
 fix(initbbs): adjust board order to let bid=1 group exist

holishing:
The group whose bid is 1 should be root of `(C)lass 分組討論區`.
Besides, the creation order of these boards determine their bid number

This patch will allow contributors to easier organize boards and groups when creating their new site
and also make pttweb able to find the entry by the route (`/cls/1`) for their new PttBBS site

### DIFF
--- a/util/initbbs.c
+++ b/util/initbbs.c
@@ -104,12 +104,19 @@ static void initBoards() {
     
     if(fp) {
 	memset(&b, 0, sizeof(b));
-	
+
+	strcpy(b.brdname, "0ClassRoot");
+	strcpy(b.title, ".... Σ分組討論區");
+	b.brdattr = BRD_GROUPBOARD;
+	b.level = PERM_SYSOP;
+	b.gid = 3;
+	newboard(fp, &b);
+
 	strcpy(b.brdname, "SYSOP");
 	strcpy(b.title, "嘰哩 ◎站長好!");
 	b.brdattr = BRD_POSTMASK;
 	b.level = 0;
-	b.gid = 2;
+	b.gid = 3;
 	newboard(fp, &b);
 
 	strcpy(b.brdname, "1...........");
@@ -123,14 +130,14 @@ static void initBoards() {
 	strcpy(b.title, "發電 ◎雜七雜八的垃圾");
 	b.brdattr = 0;
 	b.level = PERM_SYSOP;
-	b.gid = 2;
+	b.gid = 3;
 	newboard(fp, &b);
 	
 	strcpy(b.brdname, "Security");
 	strcpy(b.title, "發電 ◎站內系統安全");
 	b.brdattr = 0;
 	b.level = PERM_SYSOP;
-	b.gid = 2;
+	b.gid = 3;
 	newboard(fp, &b);
 	
 	strcpy(b.brdname, "2...........");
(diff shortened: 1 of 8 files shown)


util/initbbs.c
```diff
@@ -109,7 +109,21 @@ static void initBoards() {
 	strcpy(b.title, ".... �U��հQ�װ�");
 	b.brdattr = BRD_GROUPBOARD;
 	b.level = PERM_SYSOP;
-	b.gid = 3;
+	b.gid = 2;
+	newboard(fp, &b);
+
+	strcpy(b.brdname, "0_Group");
```
IepIweidieng, line 115: The board creation order is changed again.

Is there a way to rearrange the commits so that there is only one commit which rearranges the board creation order?

util/initbbs.c
```diff
@@ -119,90 +133,76 @@ static void initBoards() {
 	b.gid = 3;
 	newboard(fp, &b);
 
-	strcpy(b.brdname, "1...........");
```
IepIweidieng: The `boardheader_t::brdname` is changed, so this commit consists of two kinds of changes.

Although it seems that `boardheader_t::brdname` is not displayed but for sorting only, so the changes in this commit might not affect the apparent behavior (apparent to whom?).

This is different from the previous commit… or isn't it?

IepIweidieng (commented):
I produced several reviews about the arrangement of the commits.

chhsiao1981:
I thought initbbs is more like the default init settings for pttbbs (and as indication of how current pttbbs board structure is organized).

I think we can have some other customized initbbs, but maybe it's better that we still have this default initbbs, and have the other customized initbbs with some other naming (perhaps _postfix, like: initbbs_holishing)

IepIweidieng:
As an non-crew of PTT, I am not sure whether the current default is still consistent with the PTT site. (The default seems to be untouched for 19+ years)

A question is whether BID 1 should be assigned to the board `SYSOP` (as the current default) or to the root catalog group (as proposed). (Is catalog groups exist 25 years ago? Why BID 0 did not be used?)

As for the suffix idea, I think that it should be the other way around, i.e., not `initbbs.c`/`initbbs_holising.c` but `initbbs_ptt.c`/`initbbs.c`.
It might be understandable for the maintainers of future PttBBS-based sites to use a more reasonable default.

Besides, a way to avoid code duplication might be necessarily, e.g., the suffixed version just invokes the non-suffixed one with specified arguments.

chhsiao1981:
hi @IepIweidieng and @holishing ～

I am not a crew of PTT either, nor has the right to merge / reject the PR～

hungte (changes requested):
> I thought initbbs is more like the default init settings for pttbbs (and as indication of how current pttbbs board structure is organized).

The PTT1 is a very special site, so the initbbs and most of the default config are actually designed to work with a "normal" BBS, or something like PTT2.

> I think we can have some other customized initbbs, but maybe it's better that we still have this default initbbs, and have the other customized initbbs with some other naming (perhaps _postfix, like: initbbs_holishing)

Definitely not. I think that can be configured in a different way, for example by moving the boards and permissions to an external file (or in pttbbs.conf). Let's not messing the source repo.

> A question is whether BID 1 should be assigned to the board SYSOP (as the current default)

Historically it should, but I also can't find any code relying on BID=1. On the contrary, web /cls/1 is hard-coded.

> Why BID 0 did not be used?)

0 is used to indicate "error" or "no such board" in get_bnum so we can't use it.

> or to the root catalog group (as proposed). (Is catalog groups exist 25 years ago?

I think we can solve the problem in either way: (1) allow the web to support changing root class to a different bid, or (2) revise the init board generation so it's taking an extra input file instead of hard-coded in source.

No matter what, this commit must be changed. We should not make the code more messy.

holishing:
Due to recent major code updates and the fact that the method for generating initial boards is still pending confirmation.  I am going to close this pull request for now. If there are better solutions or a need for further discussion, we can open a new issue or PR to track it~

Issue #58, opened 2019-03-14
Modernize code


wens:
Some ideas to modernize the code:

  - Use arc4random and co.
  - Re-align structures and drop the packed attribute
  - Use crypt_r w/ SHA256 or SHA512
  - IPv6 support?

mshockwave:
Just as a follow-up comment:
Though it's not so straight forward, but **we definitely need testings**

hungte:
I think we're probably all done (or, nothing to do) here?

1. arc4random : already used.
2. Re-align structures and drop the packed attribute: partially, but for C network stream I see no reason to "drop pack".
3. crypt_r is replaced by fcrypt (blow fish) so no need to do SHA256/512 etc.
4. IPv6 seems already supported (or not needed) since mbbsd now only takes input from the sshd or ws?

wens:
Our production environment doesn't have IPv6, so for us it's not needed. However I am sure there are address (string or binary) fields that are not large enough for actual IPv6 usage.

PR #118, opened 2022-07-07
Unit test

chhsiao1981:
This PR intended to add unit-test with googletest and github actions.

The goal is with the capability that all the PRs can be automatically tested through github actions.

This PR is with the following features:
* most of the original dev code / deploy system (make) are not modified.
    * The modified codes can be easily check by `git diff master --name-status|grep '^M'`
    * One modified code is having main in a separated file and as the wrapper of the original main. This is to enable the tests include all the rest of the code.
    * The other modified code is 1 inline function changed to static inline function (in fav.c) because of newer definition of inline.
    * Enough examples demonstrating the feasibility of unit-tests.
    * With mocked io.
    * With mocked getrandom.
* Using bazel as the underlying testing platform, because googletest is compatible only with cmake and bazel. I think bazel is better than cmake in the long run.
* The commit is separated with several meaningful sub-commits. All the comments in the sub-commits follow the commit format.

https://github.com/ptt/pttbbs/issues/58


IepIweidieng:
Some personal thoughts from a non-maintainer:
- > `git diff master --name-status|grep '^M'`
    - `git diff origin/master --stat` also does the job.
- For future comparisons of `testutil/mock_io.c` against `mbbsd/io.c`, `diff -U 2 --color -w testutil/mock_io.c mbbsd/io.c` might do the job.
- `inline` (C99) in ISO C has always meant the same (the differences come from GNUC90 and ISO C++).
- The commit that re-indents `mock_io.c` could be optional.
    - `mock_io.c` is not test code but utillity code.
    - I think that complex codes in 2-space K&R are inconvenient for reading (although test codes are assumed to be simple) while codes in 2-space Allman with forced braces or 4-space K&R are fine.
- For the format of commit subject, I found the comment from @‌wens for reference: <https://github.com/ptt/pttbbs/pull/81#issuecomment-663349618>. I currently follow an [Angular](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#commit)-like convention.

chhsiao1981:
@IepIweidieng 

Thanks for the comments!

For the indent issue, `mock_io.c` was automatically done by [my emacs settings](https://github.com/google/styleguide/blob/gh-pages/google-c-style.el). 
I thought that many people just follow google-style guide these days?
(It seems like the indent-style is only consistent within same file in this repo?)

I myself prefer K&R style (and always with brackets). 2-space or 4-space are both good to me.

I'll refine the commit subjects.

chhsiao1981:

PR #119, opened 2022-07-13
googletest with github action

chhsiao1981:
This is the subset of #118.

The purpose of this PR is to start having a working CI.

As mentioned in #118, this PR intended to use googletest as unit-test in CI.
googletest is only compatible with cmake and bazel.
I choose to use bazel as the underlying unit-test environment.

We still can use the original make for production deployment.

The only modified files are Makefile (to include bazel.mk)
and mbbsd/fav.c (to comply with the ISO C inline definition.)


chhsiao1981:

PR #120, opened 2022-08-13
pfterm: fix broken scrolling when term height > ft.rows 修正終端機高度 > ft.rows 時，畫面捲動壞掉的問題

IepIweidieng:
`fterm_rawscroll()` required that term height == `ft.rows` in order to work properly, but it is not the case when `ft.rows` is a clipped value of term height.\
`fterm_rawscroll()` 原本要求終端機的高度 == `ft.rows` 以正確運作，但此前提在 `ft.rows` 是終端機高度的被限制値時不成立。

`fterm_rawscroll()` now only assumes the term height to be <= 9999.\
現在 `fterm_rawscroll()` 只假設終端機高度 <= 9999。

### Demonstration of the broken scrolling behavior (terminal size: 102x202):
壞掉的捲動行爲示例（終端機大小：102x102）：

https://user-images.githubusercontent.com/37586669/184506762-1a6a088e-d595-47d2-a2a1-94fc460a1310.mp4

- "Up-scrolling" (forward): The bottom line is duplicated.\
  「上捲」（向之後內容捲／向前捲）：畫面底部的行被重複。
- "Down-scrolling" (backward): The screen is not scrolled and only a line near the bottom is changed (and the screen becomes de-synchronized with the internal state of pfterm).\
  「下捲」（向先前內容捲）：畫面不捲動，且只有靠近畫面底部的行有被更新（且畫面與 pfterm 內部狀態變得不同步）。

**Note: The definition of up/down-scrolling in `pfterm.c` is opposite to the usual definition.**
**注意：`pfterm.c` 中，上／下捲的定義與通常定義相反。**

(In the usual definition, the "up/down" in "up/down-scrolling" refers to the moving direction of the screen range, not the moving direction of the screen content. However, the "up/down-scrolling" in pfterm is defined by the latter.)\
（在通常定義中，「上／下捲」的「上／下」指的是畫面範圍的移動方向，不是畫面內容的移動方向。但 pfterm 的「上／下捲」以後者定義。）

This phenomenon occurs when the terminal height > 100. (100 is the limit of terminal height of Ptt)\
此現象在終端機高度 > 100 時發生。（100 是 Ptt 的終端機高度限制）

### DIFF
--- a/mbbsd/pfterm.c
+++ b/mbbsd/pfterm.c
@@ -2187,9 +2187,17 @@ fterm_rawscroll (int dy)
     // we are not going to preserve (rx,ry)
     // so don't use fterm_move*.
     if (dy > 0)
-        fterm_rawcmd2(ft.rows, 1, 1, 'H');
+    {
+        // use a large row number in case terminal height > ft.rows
+        // The terminal height passed to resizeterm() can be unreliable
+        fterm_rawcmd2(9999, 1, 1, 'H');
+    }
     else
+    {
+        fterm_rawcmd2(ft.rows + 1 - ady, 1, 1, 'H');
+        fterm_raws(ESC_STR "[J");
         fterm_rawcmd2(1, 1, 1, 'H');
+    }
 
     for (; ady > 0; ady--)
     {


IepIweidieng:
Updates of the force-pushes:
- rework the up-scrolling fix to keep the implementation of "`ESC` `M` at `(0, x)` as `rscoll()`".
- fix incorrect new comments ("scrolling down" (forward) is referred as "scrolling up" in pfterm)

mbbsd/pfterm.c
```diff
@@ -2186,10 +2186,16 @@ fterm_rawscroll (int dy)
 
     // we are not going to preserve (rx,ry)
     // so don't use fterm_move*.
-    if (dy > 0)
-        fterm_rawcmd2(ft.rows, 1, 1, 'H');
-    else
+    if (dy > 0) {
+        // use a large row number in case terminal height > ft.row
+        fterm_rawcmd2(9999, 1, 1, 'H');
```
wens, line 2193: I think it would make more sense to save the actual terminal height in `resizeterm()` and use that here.

mbbsd/pfterm.c
```diff
@@ -2186,10 +2186,16 @@ fterm_rawscroll (int dy)
...
+    } else {
+        // clear ady rows from screen bottom before scrolling up
+        // to prevent extra rows when terminal height > ft.row
```
wens: and instead of having a comment, you could just write a conditional block here.

mbbsd/pfterm.c
IepIweidieng: The actual terminal height is not available to `resizeterm()`. The size arguments passed to `resizeterm()` are already modified by `mbbsd/term.c` `term_resize()`.\
眞實終端機高度並不爲 `resizeterm()` 所知。傳入 `resizeterm()` 的眞實大小引數已遭 `mbbsd/term.c` `term_resize()` 變更。

<https://github.com/ptt/pttbbs/blob/14a90b7bb44f5b1e0b349403428d84f2c0d8779e/mbbsd/term.c#L48-L56>
<https://github.com/ptt/pttbbs/blob/14a90b7bb44f5b1e0b349403428d84f2c0d8779e/mbbsd/term.c#L60-L61>

Calling path 呼叫路徑:
* Telnet (& bbs-sshd SSH):
    * `common/sys/telnet.c` `telnet_handler()` `ctx->callback->term_resize()` (`mbbsd/telnet.c` `telnet_cb_resize_term()`) → `mbbsd/term.c` `term_resize()`
* Former SSH 先前的 SSH: `mbbsd/term.c` `sig_term_resize()` → `mbbsd/term.c` `term_resize()`
* Common 共通: `mbbsd/term.c` `term_resize()` →  → `mbbsd/pfterm.c` (or `mbbsd/screen.c`) `resizeterm.c`

Also, pfterm is adopted for non-PttBBS BBSes. `pfterm.c` was made easily portable to a MapleBBS 3&ndash;derived BBS. The `9999`-hardcoding method still works on legacy codes.\
另外，pfterm 已受非 PttBBS 的 BBS 系統所採用。`pfterm.c` 曾經歷使其更易移植至 MapleBBS 3 衍生系統的演變。而寫死 `9999` 的方式仍能適用於既存的程式碼。

IepIweidieng:
Updates of the force-push:
* fix a compilation-failing typo: `ft.row` -> `ft.rows`
* explain why `9999` instead of a dedicated variable is used

Ready for re-review.

mbbsd/pfterm.c
wens: Ah, OK, I see it.

Can you drop the "fix(...)" from the subject? It should just say "pfterm: fix ......". Then I'll merge it.

Thanks.

IepIweidieng: Done. ~~angular obsession~~

PR #121, opened 2022-09-01
mbbsd: Ensure setting up contact-email

chhsiao1981:
contact-email will be frequently used
for important operations (resetting password / notifications)

We would like to ensure that valid contact-email
exists right after successful login.

Return err in change_contact_email.
The program exits with err if unable to set contact-email
right after successful login.


chhsiao1981:
#113

mbbsd/register.c
```diff
@@ -1158,6 +1158,18 @@ void
 check_register(void)
 {
     char fn[PATHLEN];
+    int ret = 0;
+
+    // �T�{�@�w�n���p���H�c
+    sprintf(fn, "etc/ensurecontactemail");
+    if(dashf(fn) && !is_valid_email(cuser.email)) {
```
wens: This should be determined at compile time, not based on whether some file exists or not, which introduces system call overhead on each login.

This is no longer your old session based BBS. Many apps log in and out rapidly.

include/proto.h
```diff
@@ -151,7 +151,7 @@ const char *banned_msg(const char *bname);
 
 /* cal */
 const char* money_level(int money);
-int pay(int money, const char *item, ...) GCC_CHECK_FORMAT(2,3); 
+int pay(int money, const char *item, ...) GCC_CHECK_FORMAT(2,3);
```
wens: Please drop all unrelated white space changes.

mbbsd/register.c
```diff
@@ -1158,6 +1158,18 @@ void
...
+        more(fn, NA);
+        pressanykey();
+        ret = change_contact_email();
+        if(ret) {
+          exit(1);
+        }
```
wens: Why not just do a while loop until it succeeds?

wens (changes requested):

mbbsd/register.c
```diff
@@ -1158,6 +1158,16 @@ void
 check_register(void)
 {
     char fn[PATHLEN];
+    int ret = 0;
+
+    // �T�{�@�w�n���p���H�c
+    if(!is_valid_email(cuser.email)) {
+        more("etc/ensurecontactemail\0", NA);
```
wens: C strings are already NUL terminated. You don't need to add the NUL.

mbbsd/register.c
```diff
@@ -1158,6 +1158,16 @@ void
...
+        pressanykey();
+
+        // additional parentheses around "ret =" to please compiler
+        while((ret = change_contact_email()));
```
wens: Space between `while` and parentheses.

chhsiao1981:
Hi @wens～

Thanks for the comments.
I've updated the code accordingly.

chhsiao1981:
We don't need #122 after changing the strategy by looking at the lastlogin.

#114 needs this PR to ensure valid contact email.

mbbsd/mbbsd.c
```diff
@@ -1192,6 +1194,22 @@ user_login(void)
     setproctitle("%s: %s", margs, cuser.userid);
 #endif
 
+    // ensuring the contact email if lastlogin before sometime.
+#ifdef USEREC_EMAIL_IS_CONTACT
+    #ifdef ENSURE_CONTACT_EMAIL_LASTLOGIN
+    if(cuser.lastlogin < ENSURE_CONTACT_EMAIL_LASTLOGIN) {
```
wens: No. The lastlogin check should be a separate commit.

There should be two separate commits:
a) Force user to add their contact email if the current one is invalid.
b) Force user to update their contact email if their `lastlogin` is before some timestamp.
The second item is optional and should be guarded with `#ifdef FORCE_UPDATE_CONTACT_EMAIL_LASTLOGIN`

mbbsd/mbbsd.c
```diff
@@ -1192,6 +1194,22 @@ user_login(void)
     setproctitle("%s: %s", margs, cuser.userid);
 #endif
 
+    // ensuring the contact email if lastlogin before sometime.
+#ifdef USEREC_EMAIL_IS_CONTACT
+    #ifdef ENSURE_CONTACT_EMAIL_LASTLOGIN
```
wens: `FORCE_UPDATE_CONTACT_EMAIL_LASTLOGIN` is better.

mbbsd/mbbsd.c
```diff
@@ -1192,6 +1194,22 @@ user_login(void)
...
+    if(cuser.lastlogin < ENSURE_CONTACT_EMAIL_LASTLOGIN) {
+        more("etc/ensurecontactemail", NA);
+        pressanykey();
+
+        pwcuSetEmail(empty_email);
```
wens: This isn't ideal though. If the user disconnects, they will be left with empty contact email.
You could change the underlying functions to accept a boolean parameter that allows the same email to be used. However in that case it should still be re-verified.

mbbsd/mbbsd.c
wens: Basically, you should add the force update stuff after [4b89509](https://github.com/ptt/pttbbs/commit/4b89509731a7adafab84d4d7465eae885b35867f), but with the style in that commit fixed as I mentioned before.

wens: Space between `if`/`while` and `(` please.

chhsiao1981:
@wens 

Thank you so much for taking time reviewing again.
I'll try as fast as I can to revise what you suggested.

mbbsd/mbbsd.c
chhsiao1981: > Basically, you should add the force update stuff after [4b89509](https://github.com/ptt/pttbbs/commit/4b89509731a7adafab84d4d7465eae885b35867f), but with the style in that commit fixed as I mentioned before.

Sorry for the wrong place of the code.
`lastlogin` is updated in `pwcuLoginSave` and it is after `check_register`
I'll change the code block accordingly.

chhsiao1981:
hi @wens 

Thank you so much for taking time reviewing again.
I believe that I've followed your guideline updating the code.

mbbsd/register.c
```diff
@@ -572,8 +572,25 @@ query_yn(int y, const char *msg)
...
+//     email.
+//     0: not skip checking (to check whether they are the same.)
+//     1: skip checking (not to check whether they are the same.)
+//
+// Return:
+//     0 (REGISTER_OK): ok
+//     < 0: error code
 static int
-register_email_verification(email_input_t *ein)
+register_email_verification(email_input_t *ein, int is_skip_check_same_email)
```
wens: Use `bool`. Parameter should be named `skip_same_email_check`.

sample/etc/ensurevalidcontactemail
```diff
@@ -1,4 +1,4 @@
-�z���p���H�c�w�Q���m�A
+�z���p���H�c���O���Ī��H�c�A
```
wens, line 1: Squash this change, both the text change and rename, into the previous commit.

mbbsd/register.c
```diff
@@ -1712,18 +1737,25 @@ static int notify_email_change(const char *userid, const char *email)
...
+//     whether to skip checking the input email is the same as the original
+//     email.
+//     0: not skip checking (to check whether they are the same.)
+//     1: skip checking (not to check whether they are the same.)
+//
 // Return:
 //   int: 0: ok -1: err
 int
-change_contact_email()
+change_contact_email(int is_skip_check_same_email)
```
wens: Use `bool`. Parameter should be named `skip_same_email_check`.

mbbsd/register.c
```diff
@@ -1162,11 +1179,19 @@ check_register(void)
 
     // �T�{�@�w�n���p���H�c
     if (!is_valid_email(cuser.email)) {
-        more("etc/ensurecontactemail", NA);
+        more("etc/ensurevalidcontactemail", NA);
```
wens, line 1182: Squash this into the first commit.

mbbsd/register.c
```diff
@@ -1162,11 +1179,19 @@ check_register(void)
...
         pressanykey();
 
         // additional parentheses around "ret =" to please compiler
-        while( (ret = change_contact_email()) );
+        while ((ret = change_contact_email(0)));
+#ifdef FORCE_UPDATE_CONTACT_EMAIL_LASTLOGIN
+    } else if (cuser.lastlogin < FORCE_UPDATE_CONTACT_EMAIL_LASTLOGIN) {
```
wens: `cuser.lastlogin` is already updated [here in `mbbsd.c:user_login()`](https://github.com/ptt/pttbbs/blob/master/mbbsd/mbbsd.c#L1197) before `check_register()` is called in the same function.

Instead, you should use `last_login_time`, which is a backup of `cuser.lastlogin` saved early in `start_client()` before `user_login()` is called.

wens:
The commit subject for the second commit should be different, and should read something like "implement force updating of contact email" or something.

wens (changes requested):

mbbsd/register.c
chhsiao1981: It looks like [here in `mbbsd.c:user_login()`](https://github.com/ptt/pttbbs/blob/master/mbbsd/mbbsd.c#L1197)
actually only sets the human-readable format of `cuser.lastlogin` to `lasttime`, but not changes `cuser.lastlogin`
([https://github.com/ptt/pttbbs/blob/master/common/sys/time.c#L153](https://github.com/ptt/pttbbs/blob/master/common/sys/time.c#L153),
 [https://linux.die.net/man/3/localtime_r](https://linux.die.net/man/3/localtime_r)]

However, it looks like [the convention in the code is using `last_login_time`](https://github.com/ptt/pttbbs/blob/master/mbbsd/mbbsd.c#L1269)

I'll follow the convention accordingly.

chhsiao1981:
hi @wens ～

Thank you so much for the review and comments again!
I believe that I've followed all you requests～

Can you review it again？～

Thanks!

PR #123, opened 2022-09-25
util: fix inappropriate semicolon in expire.c

chhsiao1981:
Based on c95077384, we need to remove this inappropriate semicolon.


wens:
Thanks. Squashed in and force pushed.

PR #127, opened 2023-04-22
mbbsd: timestamp in verifydb_change_userid

chhsiao1981:
This seems to be a typo.

With the existing implementation, new timestamp is always generated if the userid is changed.
This PR can be updated to be as 0 to make this policy consistent.

### DIFF
--- a/mbbsd/verifydb.cc
+++ b/mbbsd/verifydb.cc
@@ -203,7 +203,7 @@ int verifydb_change_userid(const char *from_userid, const char *to_userid,
       // (by using an empty vkey). Also, make sure the insert is successful
       // before deleting.
       if (verifydb_set(to_userid, entry->generation(), entry->vmethod(),
-                       entry->vkey()->c_str(), entry->timestamp() < 0) ||
+                       entry->vkey()->c_str(), entry->timestamp()) < 0 ||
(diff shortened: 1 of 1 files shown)


IepIweidieng (commented):
The commit message doesn't mention what happened to the timestamp in `verifydb_change_userid` and what is changed.

IepIweidieng:
* I assume `entry->timestamp()` to be a UNIX timestamp. If that's the case, `entry->timestamp() < 0` will be expected to be always `false` (type: `bool`).
* Consider that the 5th parameter `timestamp` of `verifydb_set()` is of type `int64_t` (semantically `time64_t` or the like). It's doubtful whether using a `bool` (becomes either `0` or `1` when casted into `int64_t`) as the argument is intended.
* `verifydb_set()` returns a negative value (`-1`) on failure. Such negative-failure return values are commonly compared using `< 0` (as the next occurrence of `verifydb_set()`) rather than using an implicit `!= 0`.

robertabcd:
Thanks. It is a typo. It was meant to carry over the old entry into the new entry. `verifydb_set()` returns either `0` or `-1` so it worked fine.

Can you change the commit subject to something like this?
`mbbsd: Fix timestamp on new entry in verifydb_change_userid`
The commit message should at least convey it's fixing an issue.

PR #129, opened 2023-05-17
fix pfterm fterm_rawscroll() output incompatibility by passing true term size from term_resize() 修正 pfterm fterm_rawscroll() 輸出之不相容，透過從 term_resize() 給入眞實終端機大小

IepIweidieng:
Some third-party apps assume that the cursor-moving escapes have in-bound coord, which had become untrue due to [1847f5b9d7](https://github.com/ptt/pttbbs/commit/1847f5b9d756dad5b44d308054a18ce4f93f65a9) (ptt/pttbbs#120).\
一些第三方應用程式假設游標移動控制碼的座標在畫面內，此假設由於 [1847f5b9d7](https://github.com/ptt/pttbbs/commit/1847f5b9d756dad5b44d308054a18ce4f93f65a9) (ptt/pttbbs#120) 而不再成立。

To address this, a new function `resizeterm_within()` is introduced to the API of pfterm to allow passing the full terminal size along with the display region size. \
爲解決此問題，pfterm 的應用程式介面引入了新函式 `resizeterm_within()` 以讓實際終端機大小能同時與顯示範圍大小傳入。

This partially reverts the forward scrolling handling of [1847f5b9d7](https://github.com/IepIweidieng/pttbbs/commit/1847f5b9d756dad5b44d308054a18ce4f93f65a9). \
這部份地反轉了 [1847f5b9d7](https://github.com/IepIweidieng/pttbbs/commit/1847f5b9d756dad5b44d308054a18ce4f93f65a9) 中的向後捲動處理。

See the commit message for details.\
細節請見提交訊息。

If this issue is not fixed on the server side, third-party apps will need to treat the coordinates exceeding the bottom-most row as the bottom-most row to behave correctly.\
若伺服器方未修正，第三方 app 須將超出畫面最底行的座標視爲畫面最底行以正確運作。

Relevant reports from sub-sites of PTT:\
批踢踢各分站內有關此問題的相關回報：

* `[請問] PTT star 某些文章自動捲到頁尾` [#1aP984sy (EZsoft)](https://www.ptt.cc/bbs/EZsoft/M.1684312580.A.DBC.html) (PTT Star)
* `[推薦] PTT Star瀏覽器` [#1SASLqsL (EZsoft)](https://www.ptt.cc/bbs/EZsoft/M.1546241396.A.D95.html) 推文
* `[問題] 有沒有JPTT使用者覺得今天怪怪的` [#1aOw9-ks (MobileComm)](https://www.ptt.cc/bbs/MobileComm/M.1684251262.A.BB6.html) (disconnect; JPTT)
* `R: [問題] 有沒有JPTT使用者覺得今天怪怪的` [#1aP5dq8Z (MobileComm)](https://www.ptt.cc/bbs/MobileComm/M.1684298228.A.223.html)
* `[問題] 伺服器回傳錯誤的座標位置 esc[9999;1H` [#1aP5_o5- (SYSOP)](https://www.ptt.cc/bbs/SYSOP/M.1684299762.A.17E.html)
* `[問題] JPTT無故斷線` [#1aP5Dygp (AppsForBBS)](https://www.ptt.cc/bbs/AppsForBBS/M.1684296572.A.AB3.html) (disconnect; JPTT)
* `[問題] MoPtt是不是有問題？` [#1aPBt2Us (iOS)](https://www.ptt.cc/bbs/iOS/M.1684323778.A.7B6.html) (crash; Mo PTT)
* `[系統] PTT系統連線卡卡?` #1aOvxv4- (SYSOP) \[[ptt2.cc](https://term.ptt2.cc)]

There are over 41 relevant reports from the board Gossiping (until 2023-05-18 20:00 UTC+8). Please refer to the following note for the issue state of third-party apps for details:\
Gossiping 板有 41 則以上相關報告（截至 2023-05-18 20:00 UTC+8），請見對於第三方 app 狀況的詳細整理：<https://hackmd.io/@IID/pfterm-scroll-incompat>

### DIFF
--- a/docs/pfterm.txt
+++ b/docs/pfterm.txt
@@ -6,8 +6,9 @@
     API Manual
     函式說明手冊
 
-    VERSION 1.1
-    最後更新: 2007/12/24 18:00   piaip
+    VERSION 1.2
+    v1.1:     2007/12/24 18:00   piaip
+    最後更新: 2023/06/01 01:55   IID
 
 =============================================================================
 
@@ -32,8 +33,15 @@
 
 // initialization   初始化
 void    initscr     (void);                 初始系統
-int     resizeterm  (int rows, int cols);   調整視窗大小為(rows,col)
+int     resizeterm  (int rows, int cols);   調整畫面大小為(rows,cols)
+int     resizeterm_within(                  調整顯示範圍為終端機畫面的部分區域：
+            int rows, int cols,             顯示範圍大小為(rows,cols)，
+            int rows_full, int cols_full);  實際終端機則為(rows_full,cols_full)
 int     endwin      (void);                 結束系統
+
+            resizeterm() 假設顯示範圍大小 (rows,cols) 即為實際終端機大小。
+            但當顯示範圍與實際終端機大小不符時，畫面捲動行為將會不正確；
+            可用 resizeterm_within() 指定實際終端機大小 (rows_full,cols_full)。
 
 // cursor           游標
 void    getyx       (int *y, int *x);       取得目前游標位置
--- a/include/proto.h
+++ b/include/proto.h
@@ -595,6 +595,7 @@ ChessInfo* reversi_replay(FILE* fp);
 /* screen/pfterm (ncurses-like) */
 void initscr	(void);
 int  resizeterm	(int rows, int cols);
+int  resizeterm_within(int rows, int cols, int rows_full, int cols_full);
 void getyx	(int *py, int *px);
 void move	(int y, int x);
 void clear	(void);
--- a/mbbsd/pfterm.c
+++ b/mbbsd/pfterm.c
@@ -241,7 +241,8 @@ typedef struct
     ftchar  *dmap;          // dirty map
     ftchar  *dcmap;         // processed display map
     ftattr  attr;
-    int     rows, cols;
+    int     rows, cols;     // the (possibly cropped) display region size
+    int     rows_full, cols_full; // the full terminal size
     int     y, x;
     int     sy,sx;  // stored cursor
     int     mi;     // map index, mi = current map and (1-mi) = old map
@@ -346,6 +347,7 @@ static FlatTerm ft;
 // initialization
 void    initscr     (void);
 int     resizeterm  (int rows, int cols);
+int     resizeterm_within(int rows, int cols, int rows_full, int cols_full);
 int     endwin      (void);
 
 // attributes
@@ -498,7 +500,17 @@ endwin(void)
 int
 resizeterm(int rows, int cols)
 {
+    // assume that the full terminal size == the display region size
+    return resizeterm_within(rows, cols, rows, cols);
+}
+
+int
+resizeterm_within(int rows, int cols, int rows_full, int cols_full)
+{
     int dirty = 0, mi = 0, i = 0;
+
+    ft.rows_full = rows_full;
+    ft.cols_full = cols_full;
 
     rows = ranged(rows, FTSZ_MIN_ROW, FTSZ_MAX_ROW);
     cols = ranged(cols, FTSZ_MIN_COL, FTSZ_MAX_COL);
@@ -2188,9 +2200,7 @@ fterm_rawscroll (int dy)
     // so don't use fterm_move*.
     if (dy > 0)
     {
-        // use a large row number in case terminal height > ft.rows
-        // The terminal height passed to resizeterm() can be unreliable
-        fterm_rawcmd2(9999, 1, 1, 'H');
+        fterm_rawcmd2(ft.rows_full, 1, 1, 'H');
     }
     else
     {
--- a/mbbsd/term.c
+++ b/mbbsd/term.c
@@ -52,18 +52,17 @@ void term_resize(int w, int h)
 
 
     /* make sure reasonable size */
-    h = MAX(24, MIN(100, h));
-    w = MAX(80, MIN(200, w));
+    int h_crop = MAX(24, MIN(100, h));
+    int w_crop = MAX(80, MIN(200, w));
 
-    if (w != t_columns || h != t_lines)
+    // invoke terminal system resize
+    resizeterm_within(h_crop, w_crop, h, w);
+    if (w_crop != t_columns || h_crop != t_lines)
     {
-	// invoke terminal system resize
-	resizeterm(h, w);
-
-	t_lines = h;
-	t_columns = w;
 	dorefresh = 1;
     }
+    t_lines = h_crop;
+    t_columns = w_crop;
     b_lines = t_lines - 1;
     p_lines = t_lines - 4;
 
--- a/mbbsd/screen.c
+++ b/mbbsd/screen.c
@@ -79,6 +79,16 @@ initscr(void)
 }
 
 int
+resizeterm_within(int rows, int cols, int rows_full GCC_UNUSED, int cols_full GCC_UNUSED)
+{
+    // FIXME: rows_full instead of b_lines should be used for forward scrolling
+    if (rows != t_lines || cols != t_columns) {
+        return resizeterm(rows, cols);
+    }
+    return 0;
+}
+
+int
 resizeterm(int w, int h)
 {
     screenline_t   *new_picture;


IepIweidieng:
The solution of passing true term height from logind is working in process.\
讓 logind 傳送眞實終端機高度的方案正硏究中。

IepIweidieng:
> The solution of passing true term height from logind is working in process.\
> 讓 logind 傳送眞實終端機高度的方案正硏究中。

應無必要。經確認，`start_service()` 僅在發現 `login_ctx` 從 Telnet 協定取得的畫面高或寬大於 24×80 時更新 `login_data` 的長或寬，僅當畫面高度小於 24 會造成 `t_line_true` 不正確，而此時：

- 畫面繪製：由於 `t_line_true` 大於實際値，`fterm_rawscroll()` 仍能移至畫面底部而正確運作。
- App 相容：多數第三方 app 僅解析 BBS 畫面，而不設定畫面大小或僅加高畫面高度，無此問題。

IepIweidieng:
批踢踢各分站內有關此問題的相關回報：

* `[請問] PTT star 某些文章自動捲到頁尾` [#1aP984sy (EZsoft)](https://www.ptt.cc/bbs/EZsoft/M.1684312580.A.DBC.html) (PTT Star)
* `[推薦] PTT Star瀏覽器` [#1SASLqsL (EZsoft)](https://www.ptt.cc/bbs/EZsoft/M.1546241396.A.D95.html) 推文
* `[問題] 有沒有JPTT使用者覺得今天怪怪的` [#1aOw9-ks (MobileComm)](https://www.ptt.cc/bbs/MobileComm/M.1684251262.A.BB6.html) (crash; JPTT)
* `R: [問題] 有沒有JPTT使用者覺得今天怪怪的` [#1aP5dq8Z (MobileComm)](https://www.ptt.cc/bbs/MobileComm/M.1684298228.A.223.html)
* `[問題] 伺服器回傳錯誤的座標位置 esc[9999;1H` [#1aP5_o5- (SYSOP)](https://www.ptt.cc/bbs/SYSOP/M.1684299762.A.17E.html)
* `[問題] JPTT無故斷線` [#1aP5Dygp (AppsForBBS)](https://www.ptt.cc/bbs/AppsForBBS/M.1684296572.A.AB3.html) (crash; JPTT)
* `[問題] MoPtt是不是有問題？` [#1aPBt2Us (iOS)](https://www.ptt.cc/bbs/iOS/M.1684323778.A.7B6.html) (crash; Mo PTT)
* `[系統] PTT系統連線卡卡?` #1aOvxv4- (SYSOP) \[[ptt2.cc](https://term.ptt2.cc)]

Gossiping 板有 41 則以上相關報告（截至 2023-05-18 20:00），請見對於第三方 app 狀況的詳細整理：<https://hackmd.io/@IID/pfterm-scroll-incompat>

IepIweidieng:
I had forgotten to update the relevant comment. The commit is force-updated to address this.\
先前未更新相關註解，因此強制更新此提交。

mbbsd/pfterm.c
```diff
@@ -28,6 +29,14 @@
 # endif
 #endif
 
+#ifndef FT_TRUE_ROWS
+# ifdef TERM_TRUE_ROWS
+#  define FT_TRUE_ROWS TERM_TRUE_ROWS // use the given true term height
```
wens: This is a violation of the separation of modules.

Instead, please try to rework the `resizeterm()` API, such that we pass in the "true" screen size _and_ the preferred screen size.

IepIweidieng: Solution reworked.

I added a separate function `resizeterm_true()` to the API to keep the API backward-compatible.

IepIweidieng (commented):
A new solution has been worked out. \
新解決方案已得出。

docs/pfterm.txt
```diff
@@ -6,8 +6,9 @@
     API Manual
     �禡�����U
 
-    VERSION 1.1
-    �̫��s: 2007/12/24 18:00   piaip
+    VERSION 1.2
+    v1.1:     2007/12/24 18:00   piaip
+    �̫��s: 2023/05/27 14:55   iid
```
IepIweidieng, line 11: The documentation version has been bumped to reflect the API change. \
文件版本已更迭以反映應用程式介面的變更。

docs/pfterm.txt
```diff
@@ -6,8 +6,9 @@
     API Manual
     �禡�����U
 
-    VERSION 1.1
-    �̫��s: 2007/12/24 18:00   piaip
+    VERSION 1.2
+    v1.1:     2007/12/24 18:00   piaip
+    �̫��s: 2023/05/27 14:55   IID
```
IepIweidieng: The documentation version has been bumped to reflect the API change.\
文件版本已更迭以反映應用程式介面的變更。

mbbsd/term.c
```diff
@@ -50,15 +50,21 @@ void term_resize(int w, int h)
...
     Signal(SIGWINCH, SIG_IGN);	/* Don't bother me! */
 
+#ifdef USE_PFTERM
+    // resize using the true terminal size
+    resizeterm_true(h, w);
+#endif
 
     /* make sure reasonable size */
     h = MAX(24, MIN(100, h));
     w = MAX(80, MIN(200, w));
```
IepIweidieng: This preferred screen size is ignored.

IepIweidieng: Addressed.

mbbsd/term.c
```diff
@@ -52,16 +52,22 @@ void term_resize(int w, int h)
 
 
     /* make sure reasonable size */
-    h = MAX(24, MIN(100, h));
-    w = MAX(80, MIN(200, w));
+    int h_pref = MAX(24, MIN(100, h));
+    int w_pref = MAX(80, MIN(200, w));
 
-    if (w != t_columns || h != t_lines)
+#ifdef USE_PFTERM
```
IepIweidieng: The `USE_PFTERM` here had had a typo `PFTERM`, which had prevented the fix from resolving the issue of this PR.

This typo has been fixed in the latest force-push.

include/proto.h
```diff
@@ -595,6 +595,9 @@ ChessInfo* reversi_replay(FILE* fp);
 /* screen/pfterm (ncurses-like) */
 void initscr	(void);
 int  resizeterm	(int rows, int cols);
+#ifdef USE_PFTERM
+int  resizeterm_true(int rows, int cols, int rows_true, int cols_true);
```
wens: Please add the API for both implementations. We do not want to special case for pfterm; instead switch to the new API for both. I'd keep the original resizeterm(). For `screen.c` you can just pass through to original `resizeterm()`.

I'd call the new function something like `resizeterm_with_win()`. I'd have the first two parameters as the old API, and add the "limited window" size as the latter two new parameters.

mbbsd/pfterm.c
```diff
@@ -497,9 +501,19 @@ endwin(void)
 
 int
 resizeterm(int rows, int cols)
+{
+    // assume that the true terminal size does not exceed 9999x9999
+    return resizeterm_true(rows, cols, 9999, 9999);
```
wens: No. Just assume the limited resolution is the same as the full resolution. We won't be using the old API, but we want to keep the behavior the same.

wens (changes requested):

include/proto.h
IepIweidieng: Sure. As for the new API, I had not come up with a better name for the new function. The parameters have already had the suggested semantics.

The function signature will be changed into:
```c
int resizeterm_with_win(int rows, int cols, int rows_win, int cols_win);
```

mbbsd/pfterm.c
IepIweidieng: I doubted about which behavior to keep. The pros and cons of which I thought are listed below.

I will first implement option 1 as suggested, however.

There are 3 options regarding ptt/pttbbs#120. The first 2 are incompatible with each other.

1. keep the behavior as before the forward scrolling workaround (suggested)
    * Pros:
        * It avoids the undefined effect described by [ECMA-48](https://www.ecma-international.org/wp-content/uploads/ECMA-48_5th_edition_june_1991.pdf) 6.1.7 due to out-of-bound coordinates.
        * Terminal-parsing&ndash;based 3rd-party APPs are guaranteed to work.
            * However, PTT & PTT2 are the only 2 sites which I know to have such 3rd-party APPs. Both sites will certainly use the new API and might make this irrelevant.
    * Con: BBS maintainers have to modify their code to resolve the forward scrolling issue.
        * This task may be trivial for PttBBS-based BBSes. Not necessarily so for non&ndash;PttBBS-based BBSes.
        * For PttBBS-based BBSes, it is possible that their codes either are too lagged behind PttBBS to work with 3rd-party APPs or are recent enough to simply use the new API by applying the modifications from PttBBS.
        * For non&ndash;PttBBS-based BBSes, however, I suspect that there are currently few than 10 people who are actively maintaining a non&ndash;PttBBS-based BBS which uses pfterm. Dream-Land BBS is the only site which I know to both be such a BBS and have the site currently accessible.
2. keep the behavior as introduced in the workaround (the initial choice of the `resizeterm_true()` solution of this PR)
    * Pro: BBS maintainers need only to update `pfterm.c` to resolve the issue. Using the new API is optional.
    * Con: APP maintainers may need to update their APPs for handling out-of-bound screen coordinates.
        * At this time (2023-05-29 04:20 UTC+8), almost all such APPs which I know to have been investigated either are reported to be immune or have already updated to handle it.
3. rework `fterm_rawscroll()` to make forward scrolling work independent of the actual terminal size (out of scope)
    * Pro: It avoids changing the API and does not introduce a (`struct`ed) global variable which is only used in one place.
    * It is possible to achieve this by sending Delete Line (`"\x1b[%d;M"`) per ECMA-48 and the [VT102 User Guide](https://www.vt100.net/docs/vt102-ug/chapter5.html#S5.5.2.21).
        * However, this escape is possibly not supported by some terminal emulators and terminal-parsing&ndash;based 3rd-party APPs.
        * [MapleBBS 3.00a](https://github.com/bbsmirror/BBSmirror/blob/master/Maple/Maple3/Old/Alpha/MapleBBS-3.00a.tgz) and on used Insert Line (`"\x1b[%d;L"`) for backward scrolling but fell back to the move-to-bottom-then-insert-newlines method for forward scrolling, which seems to be a refactored version of the scrolling handling of [MapleBBS 2.36b](https://github.com/bbsmirror/BBSmirror/blob/master/Maple/Maple2/MapleBBS-2.36b.tar.gz) (inherited by the non-pfterm `screen.c` of PttBBS).
        * This option needs compatibility investigation, which is out of the scope of this PR.

mbbsd/pfterm.c
```diff
@@ -2188,9 +2200,7 @@ fterm_rawscroll (int dy)
     // so don't use fterm_move*.
     if (dy > 0)
     {
-        // use a large row number in case terminal height > ft.rows
-        // The terminal height passed to resizeterm() can be unreliable
-        fterm_rawcmd2(9999, 1, 1, 'H');
+        fterm_rawcmd2(ft.rows_win, 1, 1, 'H');
```
wens: You misunderstood. `rows` / `cols` should be the full size and used here. `rows_win` / `cols_win` should be the clamped size, i.e. a cropped window.

IepIweidieng: I misunderstood because I thought the terminology *[window](https://en.wiktionary.org/wiki/window#Noun)* ("視窗") to be related to the GUI component of an OS. However, *window* can generally refer to any UI-containing rectangular area on either a terminal screen or a computer screen.

This annotated screenshot of browsing a BBS site using PCMan shows my intuition for these terminologies:
image

*[Screen](https://en.wiktionary.org/wiki/screen#Noun)* itself is an ambiguous terminology because it can refer to either the viewing area of a PC monitor ("螢幕") or a particular displayed content on such viewing area ("畫面", *e.g.,* the login screen, the [blue screen](https://en.wikipedia.org/wiki/Blue_screen_of_death)).

The primary distinction should be not between *window* & *screen* but between the total available region & the probably cropped displayed region. I think that I have not come up with simple words which cleanly distinguish between these two.

As for the variable naming, for compatibility, `ft.rows` & `ft.cols` should remain to mean the cropped display region and the newly introduced variables should be these names prefixed with a word which refers to the total available region.

wens: The API is designed to follow curses / ncurses. In that context, "terminal" means just that, the whole terminal. It does not and should not know about anything outside of the terminal, whether it be your GUI window menus and decoration, or an actual monitor. "Window" in this context means any subdivision of the terminal.

I guess you could keep using `ft.rows` and `ft.cols` internally to mean the "window" size to avoid changing all the references, and add something like `ft.rows_full` for the non-cropped size.

mbbsd/pfterm.c
```diff
@@ -242,6 +242,7 @@ typedef struct
     ftchar  *dcmap;         // processed display map
     ftattr  attr;
     int     rows, cols;
+    int     rows_win, cols_win; // the terminal window size
```
wens: Based on my other comment, this should be called `*_true`. The line above should get a comment saying "cropped or limited size".

mbbsd/pfterm.c
wens: There's only two invocations of this function. One is even inside pfterm's init function. And you are updating them both.

If someone is exporting pfterm to another application or system, that's up to them.  Either they follow curses / ncurses API definition, which this is based on, or they switch to the new custom function. We already introduced a regression anyways. Better to revert.

Regarding 3., that's not actually scrolling the terminal, and probably leads to very ugly content in the scrollback buffer. I believe pfterm was optimized to use proper scrolling to avoid needless clearing and redrawing.

BTW, the resolution cropping thing predates git history, so there's no conclusive way to say what the reason for it was. But I guess it was to avoid users randomly resizing to an insanely huge screen size, either really using it, or sending a bogus stream, and consuming huge amounts of memory for the local terminal buffer. If so then the cropping should really live in the terminal implementations (`pfterm.c` and `screen.c`). The BBS terminal/window sizes should be updated by somehow reading back the new terminal size after resizeterm() is called. ncurses provides the global variables `COLS` and `LINES` for this. Something to think about even though out of scope.

mbbsd/pfterm.c
IepIweidieng: I have renamed them instead as `*_full` as if `rows` & `cols` are always cropped for programmers to easier understand their relation to `rows` & `cols`.

IepIweidieng:
After some history investigation, I found that this exact issue was already addressed in Pirate BBS in 1990, which is the ultimate ancestor of both PttBBS and MapleBBS 3.xx.

### Pirate BBS: Uncropped `t_lines` & `t_columns` vs. Cropped `scr_lns` & `scr_cols`

In [Pirate BBS](https://github.com/bbsmirror/BBSmirror/blob/master/PirateBBS/pbbs-1.9.tar.Z) and early descendents, the BBS display region was determined by `scr_lns` & `scr_cols` and the full terminal size by `t_lines` & `t_columns`.

`scr_lns` & `scr_cols` corresponded to `ft.rows` & `ft.cols`, while `t_lines` & `t_columns` corresponded to the proposed `ft.rows_full` & `ft.cols_full`.

`scr_cols` was cropped due to the fixed size of `struct screenline`. Pirate BBS had never supported text-decorating ANSI escapes, so using a fixed buffer size for terminal texts was reasonable.

The legacy scrolling method, which inserts Insert Line (`"\x1b[;L"`) at the top row for backward scrolling and inserts newlines (`"\n"`) at the bottom row for forward scrolling, had already been employed in Pirate BBS, which used `t_lines` for the bottom row coordinate and was working even when the display region size and the terminal size differed.

In Pirate BBS, `scr_lns` & `scr_cols` was initialized to be `t_lines` & cropped `t_columns` when `initscr()` was called the first time and remained constant throughout the connection, while `t_lines` & `t_columns` could be updated only when `term_init()` was called (the resizing signal mechanism had not been employed).

### MapleBBS &ndash;2.36: Uncropped `t_lines` & May-be-cropped `t_columns` vs. Cropped `scr_lns` & Fixed `scr_cols`

In [MapleBBS 2.0.5](https://github.com/bbsmirror/BBSmirror/blob/master/Maple/Maple2/mapleBBS-2.0.5.tar.gz) (the initial public release of MapleBBS), `scr_cols` had become fixed into `ANSILINELEN`, which had effectively made it the maximum width of the display region.

Also, `t_columns` had been cropped as `scr_cols` at the first call to `initscr()`, which had affected the cursor-at-right-margin behavior. However, `t_columns` had not been cropped when updated by `term_init()`, which had caused inconsistency.

MapleBBS-SOB & PttBBS [0.00](https://www.ptt.cc/man/SetupBBS/DC13/index.html)&ndash;[0.9.2](https://github.com/bbsmirror/BBSmirror/blob/master/Ptt/openptt/pttsrc-0.9.2.tar.gz) inherited this characteristic but `big_picture` was reallocated when `t_columns` exceeded the previous maximum, so vertical resizing was potentially possible.

### MapleBBS 2.39: Only Uncropped `t_lines` & `t_columns` with Support of Horizontal Resizing

In [MapleBBS 2.39a](https://github.com/bbsmirror/BBSmirror/blob/master/Maple/Maple2/MapleBBS-2.39a.tgz), `scr_lns` & `scr_cols` were removed and replaced by `t_lines` & `t_columns`. The display region was only horizontally cropped by bound checking of struct `screenline`. However, vertical resizing was not handling properly and might have resulted in crashes due to out-of-bound accessing.

### MapleBBS 3.00: Only Fixed `t_lines` & `t_columns` with No Resizing Possible

However, in [MapleBBS 3.00b](https://github.com/bbsmirror/BBSmirror/blob/master/Maple/Maple3/Old/Alpha/MapleBBS-3.00b.tgz) (the currently known earliest complete MapleBBS 3.00 codebase), `bbsd` (previously `mbbsd`) had had only the daemon mode implemented. The `term_init()` functionality was useless and had been removed while Telnet resizing had not been implemented, which had left there only constants `t_lines` (`24`) & `t_columns` (`80`). The dynamically allocated `big_picture` had been replaced by the statically allocated `vbuf[]`.

Note: The non-daemon mode of `mbbsd` in PttBBS was a later enhancement inherited from MapleBBS-SOB, where an earlier version of `mbbsd` had been backported.

### PttBBS ((Pre-)CVS Era): Uncropped `t_lines` & May-be-cropped `t_columns` vs. Cropped `scr_lns` & Fixed `scr_cols` vs. Fixed `SCR_WIDTH`

In [PttBBS 0.9.2](https://github.com/bbsmirror/BBSmirror/blob/master/Ptt/openptt/pttsrc-0.9.2.tar.gz), the resizing signal mechanism had been employed and a constant `SCR_WIDTH` (`80`) had been introduced for the width of the display region of editable texts in the editor.

### PttBBS (SVN Era until Now): Toward Cropped `t_lines` & `t_columns` vs. Aliased Fixed `SCR_COLS` from `screen.c` or Doubly Cropped `ft.rows` & `ft.cols` from pfterm

In lantw44m/pttbbs@98a08b5b0948f96b0a431a7a1d3d420ae5a2318a ("let scr_lns = t_lines;"), `scr_lns` was made the same as `t_lines`, which has allowed the display region to resize vertically.

In lantw44m/pttbbs@809eece5367b404faf7d1d9fc1b9e9ab2cf4b7c2 ("turn on full screen support if shell login."), `t_lines` vs. `scr_lns` was commented to be confusing.

In lantw44m/pttbbs@c4c4f8e3bbeef1608f06f145a013ed66e9fd74ac ("not only long screen, but also width screen"), `SCR_WIDTH` (`80`) and some hardcoded widths were replaced with `t_columns`, which has allowed the display regions of the article browser and the editor to resize horizontally. However, `t_lines` & `t_columns` has become cropped in this commit.

In lantw44m/pttbbs@2e46e59ff7ca640b5c61b9f9d0cf57c1a6d11e60 ("t_columns should not set as ANSILINELEN"), `t_columns` was no longer cropped separately in `initscr()`, which has fixed the inconsistency issue.

In ptt/pttbbs@97289014afdc7043301d20e80b18538deb308480 ("merge back from victor.screen branch"), `scr_lns` & `scr_cols` was made into `#define`s to simplify the codes.

In ptt/pttbbs@21facac71a952db22e54be589e75ccda583a9eac ('- pfterm: piaip's flat terminal system, aka then "Perfect Term".'), pfterm was introduced with `ft.rows` & `ft.cols` cropped from cropped `t_rows` & `t_columns`, *i.e.*, doubly cropped.

Finally, in ptt/pttbbs@e12bf07b8196a9db2e2a92e2b2b7261dc93f9540 ("Remove warnings, caught by clang."), `scr_lns` was replaced by `t_lines` while `scr_cols` was renamed into `SCR_COLS`, which has leaded to the state today.

PR #132, opened 2023-12-17
util/shmctl: refine command-line UI: SHMinit → init, reworded usages

IepIweidieng:
* `shmctl SHMinit`, used to invoke `resolve_garbage()` (for pttcache, aka. ad-banner/active-board cache) & `resolve_fcache()` (for from-info cache) before doing `shmctl init` but required `shmctl init` to run beforehand, is now removed and `shmctl init` should be used instead.
* An invalid command also no longer requires `uhash_loader` to execute, so `shmctl -h` now always works.
* The options of each `shmctl` command have been made more clarified.
* The missing parameter of `shmctl init` has now been documented.

See the commit messages for details.

### DIFF
--- a/util/shmctl.c
+++ b/util/shmctl.c
@@ -100,7 +100,7 @@ int utmpfix(int argc, char **argv)
 	    upperbound = atoi(optarg);
 	    break;
 	default:
-	    printf("usage:\tshmctl\tutmpfix [-n] [-t timeout] [-F] [-D sleep]\n");
+	    printf("usage:\tshmctl\tutmpfix [-n (fast)] [-t timeout] [-l target] [-F (fork)] [-D sleep] [-u trigger]\n");
 	    return 1;
 	}
 
@@ -586,7 +586,7 @@ int setglobal(int argc, char **argv)
 {
     int     where, value;
     if( argc != 3 ){
-	puts("usage: shmctl setglobal (GV2) newvalue");
+	puts("usage: shmctl setglobal <GV2> newvalue");
 	return 1;
     }
     value = atoi(argv[2]);
@@ -846,7 +846,7 @@ int SHMinit(int argc, char **argv)
 	    no_uhash_loader = 1;
 	    break;
 	default:
-	    printf("usage:\tshmctl\tSHMinit\n");
+	    printf("usage:\tshmctl\tSHMinit [-n (no utmpsortd)]\n");
 	    return 0;
 	}
 
@@ -1163,9 +1163,9 @@ struct Cmd {
     const char    *cmd, *descript;
 } cmd[] = { 
     {dummy,      "\b\b\b\bStart daemon:", ""},
-    {utmpsortd,  "utmpsortd",  "utmp sorting daemon"},
+    {utmpsortd,  "utmpsortd",  "[interval-ms [sortxid-per-sortall]] utmp sorting daemon"},
 #ifdef NOKILLWATERBALL
-    {nkwbd,      "nkwbd",      "NOKillWaterBall daemon"},
+    {nkwbd,      "nkwbd",      "(-h for usage) NOKillWaterBall daemon"},
 #endif
 
     {dummy,      "\b\b\b\bBuild cache/fix tool:", ""},
@@ -1173,7 +1173,7 @@ struct Cmd {
     {fixbcache,  "fixbcache",  "fix bcache"},
     {rlfcache,   "reloadfcache", "reload fcache"},
     {bBMC,       "bBMC",       "build BM cache"},
-    {utmpfix,    "utmpfix",    "clear dead userlist entry & kick idle user"},
+    {utmpfix,    "utmpfix",    "(-h for usage) clear dead userlist entry & kick idle user"},
     {utmpreset,  "utmpreset",  "SHM->busystate=0"},
     {utmpwatch,  "utmpwatch",  "to see if busystate is always 1 then fix it"},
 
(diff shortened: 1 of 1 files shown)


util/shmctl.c
```diff
@@ -1226,6 +1216,18 @@ int main(int argc, char **argv)
...
+	   because uhash_loader is not run yet.      */
+	return SHMinit(argc - 1, &argv[1]);
+    }
+
+    attach_SHM();
+    /* shmctl doesn't need resolve_boards() first */
+    //resolve_boards();
+    resolve_garbage();
+    resolve_fcache();
+    return cmd[i].func(argc - 1, &argv[1]);
```
chhsiao1981, line 1232: The flow becomes:
1. if argc >= 2: find the corresponding cmd.
2. if argc == 1: print all commands and return.
3. special case for SHMinit.
4. other cmds.

Suggestion:
1. if argc == 1: print and return.
2. find the corresponding cmd.
3. special case for SHMinit.
4. other cmds.

chhsiao1981: Actually we do need to find the corresponding cmd first
because it's possible that the cmd (argv[1]) is an invalid cmd.

The original flow is good.

hungte:
need rebase

IepIweidieng:
Rebased.

util/shmctl.c
```diff
@@ -1106,6 +1096,18 @@ int main(int argc, char **argv)
...
+	/* in this case, do NOT run attach_SHM here.
+	   because uhash_loader is not run yet.      */
+	return SHMinit(argc - 1, &argv[1]);
+    }
+
+    attach_SHM();
+    /* shmctl doesn't need resolve_boards() first */
+    //resolve_boards();
+    resolve_garbage();
+    resolve_fcache();
```
hungte: I'd recommend revising this to:
````
bool do_init = false;

if (cmd[i].func == SHMinit)
   do_init = true;

if (!do_init) {
  attach_SHM();
  resolve_garbage();
  resolve_fcache();
}

return cmd[i].func(argc - 1, &argv[1]);
````

File: util/shmctl.c [hunk 3]
```diff
@@ -1071,7 +1071,8 @@ struct Cmd {
     {dummy,      "\b\b\b\bMisc:", ""},
     {showglobal, "showglobal", "show GLOBALVAR[]"},
     {setglobal,  "setglobal",  "set GLOBALVAR[]"},
-    {SHMinit,    "SHMinit",    "initialize SHM (including uhash_loader)"},
+    {SHMinit,    "init",       "initialize SHM (including uhash_loader)"},
```
hungte: Are we sure init == SHMinit?

I thought 'init' was for 'system boot up' (requires to run uhash_loader) and SHMinit is for "re-init SHM structure but the SHM was already allocated)

Although, I never figured out if they really work in this way.

util/shmctl.c
```diff
@@ -100,7 +100,7 @@ int utmpfix(int argc, char **argv)
 	    upperbound = atoi(optarg);
 	    break;
 	default:
-	    printf("usage:\tshmctl\tutmpfix [-n] [-t timeout] [-F] [-D sleep]\n");
+	    printf("usage:\tshmctl\tutmpfix [-n (fast)] [-t timeout] [-l target] [-F (fork)] [-D sleep] [-u trigger]\n");
```
hungte: Since we're here, can we make it really match for modern style, i.e.,
````
printf("usage: shmctl utmpfix <options>\n"
       "options:\n"
       "  -n: fast\n"
       "  -t <timeout>: XXX?\n"
       "  -l <target>: ...);
````

File: util/shmctl.c [hunk 5]
```diff
@@ -747,7 +747,7 @@ int SHMinit(int argc, char **argv)
 	    no_uhash_loader = 1;
 	    break;
 	default:
-	    printf("usage:\tshmctl\tSHMinit\n");
+	    printf("usage:\tshmctl\tSHMinit [-n (no utmpsortd)]\n");
```
hungte: I think it's really weird to use \t in usage - especially the setglobal above is also using only spaces.
````
"usage: shmctl SHMinit [-n]\n"
"      -n: No upmpsortd\n");
````

File: util/shmctl.c [hunk 6]
```diff
@@ -1046,32 +1046,32 @@ struct Cmd {
     const char    *cmd, *descript;
 } cmd[] = { 
     {dummy,      "\b\b\b\bStart daemon:", ""},
-    {utmpsortd,  "utmpsortd",  "utmp sorting daemon"},
```
hungte, line 1049: The format is call, cmd ,description so I think we should have a common format of description. Maybe something like
````
func. Args: [list]
````

So we have
````
"fix bcach",
"clear dead userlist entry & kick idle user. Args: [-h: for usage]"
"list board info in SHM. Args: [board-id]"
"show statistics. Args: [-c: clear stats]"

util/shmctl.c
IepIweidieng: SHMinit is for 'system boot up' (optionally including and required to run uhash_loader). Executing it twice causing errors.

hungte: I mean, I never figured out when people will run "SHMinit" (command), because that will do attachSHM and fail. "init" is more reasonable. Or, maybe "SHMinit" command actually won't fail in the past... well, I don't know how people use it previously.

If the two commands are identical, what about just keep one?

IepIweidieng: `init` is a ~~new~~ alias to `SHMinit`, ~~it never existed before this PR.~~
`init` has always been an unlisted alias to `SHMinit`.
~~`SHMinit` is kept to not break existing scripts.~~

util/shmctl.c
IepIweidieng: I had tried both way.
Putting options before connects makes it consistent to the command usage format, but putting after making the description aligned and easier to read.

util/shmctl.c
hungte: But if you search 'shmctl init' in the codebase, actually you can find it in sample/pttbbs.sh, sample/rc.local. docs/ADVANCE . Even docs/INSTALL is telling people to do "shmctl init" instead of "shmctl SHMinit".

As a result, I really don't understand why there's a SHMinit command that will first attach SHM then re-init...

util/shmctl.c
hungte: Ok got it :)

util/shmctl.c
IepIweidieng: After reviewing the history, I have finally recalled that `shmctl init` used to calls `SHMinit()` by exceptional string matching, and `shmctl SHMinit` used to break with error due to uninitialized SHM, so this PR was for fixing it and explicitly listing `init` as an existing command.

IepIweidieng: They had different behaviors, but I am not sure whether the differences are intended.
* old `shmctl init`: call `uhash_loader` to set up SHM, rebuild `bcache` (board cache) & `BMcache` (board manager cache), and start `sutmpsortd` and helper services
* old `shmctl SHMinit` (requires running `shmctl init` beforehand): `resolve_garbage()` (keep ad-banner/active-board cache updated), `resolve_fcache()` (keep from-info cache updated), and do `shmctl init`

hungte: Thanks for clarifying the history. I also searched entire PTT1 configuration and couldn't find any place that calls 'SHMinit'. IMHO we can remove that command and keep only 'init', so scripts relying on SHMinit old behavior may notice the difference (fail to run) easier, rather than getting an unexpected behavior change. Although I do suspect no one really calls "shmctl SHMinit".

BTW, looking at the history, things can go back to commit dada3acd (initial drop to svn) ,where the code was `{SHMinit,    "SHMinit",    "initialize SHM (including uhash_loader)"},`. And 'shmctl' does not exist in the ancestor OpenPTT. So unfortunately we don't have the change of how SHMinit was evolved.

I tend to believe that the "SHMinit" was created for the purpose of "init" in the very beginning; and the developer found that (maybe in some refactoring) SHMinit can't be called after attachSHM, so they added a new entry(init) and forgot to remove the SHMinit.

Issue #133, opened 2024-04-15
unable to compile boardd / mand in debian:bookworm due to libgrpc29.


chhsiao1981:
```
bbs@946d4dbea71b:~/pttbbs/daemon/boardd$ pmake clean all
rm -f boardc.o boardd.o convert.o article.o strings.o board_service_impl.o bbs++.o evbuffer.o board.pb.cc board.grpc.pb.cc board.pb.h board.grpc.pb.h board.pb.o board.grpc.pb.o boardc boardd
protoc -I. --cpp_out=. --grpc_out=. --plugin=protoc-gen-grpc=/usr/bin/grpc_cpp_plugin board.proto
ccache clang++ -g -Os -W -Wall -Wunused  -Wno-missing-field-initializers -Wno-address-of-packed-member -pipe -DBBSHOME='"/home/bbs"' -I../../include -Wno-invalid-source-encoding  -std=c++17 -I../.. -I. -Wno-unused-parameter  -c board.pb.cc
ccache clang++ -g -Os -W -Wall -Wunused  -Wno-missing-field-initializers -Wno-address-of-packed-member -pipe -DBBSHOME='"/home/bbs"' -I../../include -Wno-invalid-source-encoding  -std=c++17 -I../.. -I. -Wno-unused-parameter  -c board.grpc.pb.cc
ccache clang++ -g -Os -W -Wall -Wunused  -Wno-missing-field-initializers -Wno-address-of-packed-member -pipe -DBBSHOME='"/home/bbs"' -I../../include -Wno-invalid-source-encoding  -std=c++17 -I../.. -I. -Wno-unused-parameter  -c boardc.cpp
ccache clang++ -g -Os -W -Wall -Wunused  -Wno-missing-field-initializers -Wno-address-of-packed-member -pipe -DBBSHOME='"/home/bbs"' -I../../include -Wno-invalid-source-encoding  -std=c++17 -I../.. -I. -Wno-unused-parameter -o boardc board.pb.o board.grpc.pb.o boardc.o -Os -Wl,--as-needed  ../../util/util_var.o  ../../common/bbs/libcmbbs.a  ../../common/sys/libcmsys.a  ../../common/osdep/libosdep.a  -levent   -pthread -lstdc++  -lgrpc++ -lprotobuf -lgflags
/usr/bin/ld: board.grpc.pb.o: undefined reference to symbol 'gpr_log'
[...]

PR #134, opened 2024-07-16
daemon/*/Makefile: better grpc ldflags

chhsiao1981:
#133 

https://github.com/grpc/grpc/blob/v1.51.1/examples/cpp/helloworld/Makefile#L28
https://github.com/grpc/grpc/blob/v1.30.2/examples/cpp/helloworld/Makefile#L28


wens:
Done in commit 1ef2a1271d603d1856a9df54fa7f656e197c3a89 with same style as existing variables for `libevent`.

Issue #135, opened 2024-09-01
Unable to compile source code in Debian Trixie(13), glibc >= 2.38


holishing:
Just add a memo (from https://github.com/bbsdocker/imageptt/issues/8 ).  
If someday I find a solution, I will share here and perhaps add a PR about this.

https://github.com/bbsdocker/imageptt/actions/runs/9700365062/job/26771584227
```
#19 56.05 In file included from mergedir2.cc:7:
#19 56.05 In file included from ../include/bbs.h:13:
#19 56.05 /usr/include/string.h:506:15: error: exception specification in declaration does not match previous declaration
#19 56.05 extern size_t strlcpy (char *__restrict __dest,
#19 56.05               ^
#19 56.05 ../include/osdep.h:33:8: note: previous declaration is here
#19 56.05 size_t strlcpy(char *dst, const char *src, size_t size);
#19 56.05        ^
#19 56.05 In file included from mergedir2.cc:7:
#19 56.05 In file included from ../include/bbs.h:13:
#19 56.05 /usr/include/string.h:512:15: error: exception specification in declaration does not match previous declaration
#19 56.05 extern size_t strlcat (char *__restrict __dest,
#19 56.05               ^
#19 56.05 ../include/osdep.h:37:8: note: previous declaration is here
#19 56.05 size_t strlcat(char *dst, const char *src, size_t size);
#19 56.05        ^
#19 56.15 2 errors generated.
#19 56.16 *** Error code 1
#19 56.16 
#19 56.16 Stop.
#19 56.16 bmake[1]: stopped in /home/bbs/pttbbs/util
#19 56.16 *** Error code 1
```

IepIweidieng:
The assumption of `strlcpy()` & `strlcat()` being nonexistent on Linux is broken on your Linux distribution version.

https://github.com/ptt/pttbbs/blob/c4f6a2910ecde07f62445c38ecc980f5442f4838/include/osdep.h#L12-L24

Maybe an actual test for the existence of a certain system function is needed, which can be achieved by, *e.g.*, compiling and linking a test program without the BBS library to see whether there are any undefined references.

wens:
Please give the branch `check-for-strlcpy` a try.

wens:
Should be fixed with commit 758091331f92d78b9d2279a06c73ab50d3e718c7.

PR #136, opened 2025-11-01
make: avoid strlcpy/strlcat redefinition on glibc 2.38+

holishing:
Newer glibc (≥ 2.38) provides built-in implementations of strlcpy() and strlcat(). Projects that also define these functions will now fail to build due to symbol redefinition conflicts.

This patch is trying to verify whether strlcpy()/strlcat() are available at compile time and only compiles the compatibility implementation when missing. (e.g. glibc is older 2.38)

ref: <https://sourceware.org/git/?p=glibc.git;a=commit;h=454a20c8756c9c1d55419153255fc7692b3d2199>

related previous discussion: #135
related build test: <https://github.com/bbsdocker/imageptt/actions/runs/19000701482>

### DIFF
--- a/include/osdep.h
+++ b/include/osdep.h
@@ -18,13 +18,6 @@
 #include <sys/file.h>      /* for flock() */
 #include <strings.h>       /* for strcasecmp() */
 
-#if !defined(__has_builtin) || !__has_builtin(strlcpy)
-#define NEED_STRLCPY
-#endif
-#if !defined(__has_builtin) || !__has_builtin(strlcat)
-#define NEED_STRLCAT
-#endif
-
 #define NEED_SETPROCTITLE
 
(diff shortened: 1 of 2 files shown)


IepIweidieng:
758091331f92d78b9d2279a06c73ab50d3e718c7 failed because `__has_builtin()` only checks for functions built in the compiler, as listed in <https://gcc.gnu.org/onlinedocs/gcc/Built-in-Functions.html>. It never checks again glibc or other libraries.
In GCC, only some standard C functions have compiler built-in version. Nonstandard C library functions (here `strlcpy()` and `strlcat()`) do not have compiler built-in version.

IepIweidieng:
Resolved in 0ac8bf7038 using a glibc version check.

wens:
glibc version check seems a bit better than doing an autoconf like check.

PR #142, opened 2026-09-05
fix gcc 13.4 build failure by disabling `-Werror=format`

IepIweidieng:
* replacing `-Werror=format` -> `-Wformat`
    * gcc 13.4.0 still throws warnings due to more comprehensive checks (mostly string length checks).
* `-Wall -Wunused` with `-Wall -Wextra`
*  `-Wformat` -> `-Wformat=2`


pttbbs.mk
```diff
@@ -25,8 +25,13 @@ endif
...
 # Common build flags
 
-PTT_WARN:=	-W -Wall -Wunused -Werror=format \
+PTT_WARN:=	-W -Wall -Wunused -Wformat \
 		-Wno-missing-field-initializers -Wno-address-of-packed-member
+ifdef WARN_IS_ERROR
+PTT_WARN+=	-Werror
+else ifdef WARN_FORMAT_IS_ERROR
+PTT_WARN+=	-Werror=format
+endif
```
hungte: I think we don't need to add such of level control here. On clang, the -Werror will be applied; for gcc, given we can't enable -Werror this is useless.

pttbbs.mk
```diff
@@ -25,8 +25,13 @@ endif
 
 # Common build flags
 
-PTT_WARN:=	-W -Wall -Wunused -Werror=format \
+PTT_WARN:=	-W -Wall -Wunused -Wformat \
```
hungte: Let's make this modernized:
````
-Wall -Wextra -Wformat=2

